PR #248, opened 2014-09-09
Use dpkg_package resource to install mod_pagespeed

richid:
Using the `package` resource will default to `apt` and will error out with
`[2014-09-09T18:05:04+00:00] ERROR: package[mod_pagespeed]
(apache2::mod_pagespeed line 27) had an error:
Chef::Exceptions::Package: apt package provider cannot handle source
attribute. Use dpkg provider instead`

### DIFF
--- a/recipes/mod_pagespeed.rb
+++ b/recipes/mod_pagespeed.rb
@@ -24,7 +24,7 @@ if platform_family?('debian')
     action :create_if_missing
   end
 
-  package 'mod_pagespeed' do
+  dpkg_package 'mod_pagespeed' do
     source "#{Chef::Config[:file_cache_path]}/mod-pagespeed.deb"
     action :install
   end


whitehat101:
I ran into this bug on a fresh Ubuntu 2014 LTS. It would be nice if the bug was reopened.

I am considering the same fix as OP, simply use "dpkg_package" instead. For now, I'll just directly import the recipe/template into my cookbook.

I might spend some time getting CI to pass, since that seems to be what is holding this issue up (other than being "Closed")

This seems like a relatively simple change, does anyone remember why it is complicated?

drpebcak:
I believe @richid was having trouble running the test suites.

richid:
I spent several hours trying to get my environment set up in order the run the tests. Even with an unmodified master branch I wasn't able to have a clean build so I just made the change locally (not using Berkshelf) and called it a day.

svanzoest:
@richid I am bummed to hear you had trouble setting up the test environment. Were you using the chefdk?

jkulak:
I'm having exact same issue on Ubuntu 14.04.3 LTS (GNU/Linux 3.13.0-63-generic x86_64).

drpebcak:
@jkulak please report issues with mod_pagespeed over in the mod_pagespeed_cookbook repository. https://github.com/drpebcak/mod_pagespeed_cookbook

Issue #238, opened 2014-08-15
apache2 service only_if guard timeout exceeded: default.rb::line 36


greenreign:
**Script timed out running httpd -t**
----
```
 Mixlib::ShellOut::CommandTimeout
           --------------------------------
           Command timed out after 2s:
           Command execeded allowed execution time, process terminated
           ---- Begin output of /usr/sbin/httpd -t ----
           STDOUT:
           STDERR:
           ---- End output of /usr/sbin/httpd -t ----
           Ran /usr/sbin/httpd -t returned
```
----
When I run the script from the box it returns
Syntax OK
But it always takes about 5 seconds.  Can you make the timeout on the only_if guard longer or configurable? see default.rb::line 36

svanzoest:
The more I think about this is that what happens when the test never completes? Do we really have a case where it does not complete in 10 seconds? Ultimately there needs to be a time out somewhere. I do not really have the time to test this out, so feel free to reopen this if someone has an example and we can actually confirm what the behavior is when the test never completes.

docwhat:
I'm seeing two things that can mitigate this problem.  Doing both would be ideal.

One, `httpd -t` runs on every chef-run.  Ideally, 'httpd -t' should only be run when configurations change, not every run.  For example, for our in-house sshd configuration management, we create the config file(s) in the cache and run `sshd -T` (which is like `httpd -t`) on those cache files. If it passes, we copy over the new files using normal chef resources (e.g. so it won't do it if they didn't change). Iff the copy makes a change then it triggers a restart.

There are a lot of apache config files and it would be tricky to exactly the same trick with apache, but maybe something else could be done, just as only running `httpd -t` when `start` and `graceful` are triggered, not for all the actions on the service.

Two, `httpd -t` should never "just hang" (not including things like filesystem corruption, or the kernel crashing).  If it is taking a long time it is because the disk IO is slow, the VM is being moved across the country live, the system is really busy, or maybe swap is thrashing then even 10 seconds is too short.  The point being, in that case, chef should continue to do its job, even if it takes a while. It is up to the monitoring software to alert someone that the system is very busy.  If the next Chef run fails because chef is still running from the first time, someone will get notified then.

I would suggest if you must put a timeout, set it really high, like 3 minutes.  That way it acts as an absolutely last resort measure.

sergio-bobillier:
[...]

PR #260, opened 2014-10-10
fixing #238 issue

podwhitehawk:
https://github.com/viverae-cookbooks/apache2/issues/238 will never appear if you wait until "httpd -t" return good exit code not breaking it with timeout


svanzoest:
Ultimately there needs to be a time out somewhere. I do not really have the time to test this out, so feel free to reopen this if someone has an example and we can actually confirm what the behavior is when the test never completes.

PR #271, opened 2014-10-17
Remove fixtures not present in supported platforms

Maks3w:


svanzoest:
Thanks for this. I have removed the ones we are not using any longer.

PR #279, opened 2014-10-19
mod_ssl recipe no longer removes yum installed ssl.conf

roylindauer:


svanzoest:
As anyone had any time to test this? It would be good to get this into a release.

PR #288, opened 2014-11-05
Package install options

jespada:
This PR allows user specify package install options on RH family. Useful when you have same package on your own repo and want to install from there

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,6 +150,7 @@ attributes/default.rb file for default values in the case statement at
 the top of the file.
 
 * `node['apache']['package']` - Package name for Apache2
+* `nodet['apache']['package_options']` - Package install options for RH family
 * `node['apache']['perl_pkg']` - Package name for Perl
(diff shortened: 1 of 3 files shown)


drpebcak:
This is an interesting request.. what is stopping you from using exclude/includepkgs to control what packages you get from a repo? It seems like that is a cleaner way of ensuring that only a specific set of packages come from that repo.

jespada:
@drpebcak nothing I guess, Its just "simpler" for me to do on demand, but I probably should take a look at exclude/include.

drpebcak:
@jespada the reason I suggest it is because you can easily set that up when you configure repos, as opposed to every cookbook needing to change so you can specify custom arguments.

svanzoest:
Considering the use case is limited and would require all chef cookbooks to be adjusted. I am going to close this for now. If we find a more global/generic way, I am all ears and please file a new issue.

PR #293, opened 2014-11-19
Add a flag to allow to install mod_fastcgi from source (even on Debian familly)

jujugrrr:
You might need to install mod_fastcgi from source, even if you use Debian and have the mod_fastcgi package.
The PR is adding unit and integration tests for mod_fastcgi as well


svanzoest:
Merged. Thanks!

jujugrrr:
@svanzoest :-)

PR #299, opened 2014-12-07
FCGI paths should not be messed with on RHEL/CenOS 7.

haimgel:
CentOS 7 (and recent Fedoras) have Apache 2.4, where FCGI socket path
and shared memory path is managed adequately without further involvment
neccessary (subdirectory is created under /var/run/httpd).

However, when the path is specified manually, that subdirectory is NOT
created automatically if it doesn't already exist and Apache fails to
start. Since in recent RHEL systems /var/run is on tmpfs, end result is
that chef-created subdirectory disappears on shutdown, and Apache fails
to start after server reboot. Best to leave out these two settings
unset then.

### DIFF
--- a/recipes/mod_fcgid.rb
+++ b/recipes/mod_fcgid.rb
@@ -29,12 +29,15 @@ elsif platform_family?('rhel', 'fedora')
     backup false
   end
 
-  directory '/var/run/httpd/mod_fcgid' do
-    owner node['apache']['user']
-    group node['apache']['group']
-    recursive true
-    only_if { node['platform_version'].to_i >= 6 }
+  if ((node['platform_family'] == 'rhel') && (node['platform_version'].to_i == 6)) ||
+      ((node['platform_family'] == 'fedora') && (node['platform_version'].to_i < 20))
+    directory '/var/run/httpd/mod_fcgid' do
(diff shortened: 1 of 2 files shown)


josephholsten:
This seems reasonable, but needs to be updated onto the current code.

recipes/mod_fcgid.rb
```diff
@@ -29,12 +29,15 @@
     backup false
   end
 
-  directory '/var/run/httpd/mod_fcgid' do
-    owner node['apache']['user']
-    group node['apache']['group']
-    recursive true
-    only_if { node['platform_version'].to_i >= 6 }
+  if ((node['platform_family'] == 'rhel') && (node['platform_version'].to_i == 6)) ||
```
damacus: A library helper here called something like rhel_6? would make this statement more readable.

recipes/mod_fcgid.rb
```diff
@@ -29,12 +29,15 @@
...
+      ((node['platform_family'] == 'fedora') && (node['platform_version'].to_i < 20))
```
damacus: Same for this, something like `above_fedora_20?`

damacus:
Closing due to inactivity. 

PR #300, opened 2014-12-11
Support custom directives

brauliobo:


svanzoest:
Is `apache_conf` definitions good enough here? It seems to cover all of the cases I have encountered.

svanzoest:
I am going to close this, but please re-open if you believe that `apache_conf` does not cover your use case and we can look into this more. I am trying to keep the cookbook as simple as possible (and hopefully simplify it even more) and this adds just another way of doing things.

PR #311, opened 2015-01-23
Issue310hotfix

joshuacox:
assigning ownership of the /var/lock/apache2 directory to the apache user as per:

https://github.com/viverae-cookbooks/apache2/issues/310


drpebcak:
I've already fixed this for the specific case in my fork.

PR #314, opened 2015-02-10
parametarise AccessFileName, which defaults to .htaccess

joelhandwell:


svanzoest:
Hi,

Thanks for this. Do we want to add some chefspec/serverspec tests to ensure that the security rules work? I think it would be pretty scary if we leave a whole here.

svanzoest:
Did we ever end up writing any tests for this? I'd like to pull it in, but want to make sure we have test coverage.

joelhandwell:
@svanzoest 
chefspec: check if template generate proper config file for default access file name (.htaccess) and custom access file name (.customaccess)

serverspec: check if apache config file contain access file name 

Please review. Thank you.

svanzoest:
:+1: merged.

PR #321, opened 2015-03-03
Correct default_site name attribute in not_if guards

gsreynolds:
https://github.com/svanzoest-cookbooks/apache2/commit/4aa5d2b7bc9487dc450f70adaded1f5a83356145 introduced not_if guards to the default site file and link deletion but used `node['default_site_name']` instead of `node['apache']['default_site_name']`

### DIFF
--- a/recipes/default.rb
+++ b/recipes/default.rb
@@ -33,13 +33,13 @@ end
 %w(default default.conf 000-default 000-default.conf).each do |site|
   link "#{node['apache']['dir']}/sites-enabled/#{site}" do
     action :delete
-    not_if  { site == node['default_site_name'] }
+    not_if  { site == node['apache']['default_site_name'] }
   end
 
   file "#{node['apache']['dir']}/sites-available/#{site}" do
     action :delete
     backup false
-    not_if  { site == node['default_site_name'] }
+    not_if  { site == node['apache']['default_site_name'] }
   end
 end
 
--- a/recipes/default.rb
+++ b/recipes/default.rb
@@ -33,13 +33,13 @@ end
 %w(default default.conf 000-default 000-default.conf).each do |site|
   link "#{node['apache']['dir']}/sites-enabled/#{site}" do
     action :delete
-    not_if  { site == node['apache']['default_site_name'] }
+    not_if { site == "#{node['apache']['default_site_name']}.conf" && node['apache']['default_site_enabled'] }
   end
 
   file "#{node['apache']['dir']}/sites-available/#{site}" do
     action :delete
     backup false
-    not_if  { site == node['apache']['default_site_name'] }
+    not_if { site == "#{node['apache']['default_site_name']}.conf" && node['apache']['default_site_enabled'] }
   end
 end
 


svanzoest:
:+1: Good catch. Thanks

gsreynolds:
Just noticed that this will continuously delete the default site then recreate it as well.

```
* link[/etc/apache2/sites-enabled/default] action delete (up to date)
  * file[/etc/apache2/sites-available/default] action delete (up to date)
  * link[/etc/apache2/sites-enabled/default.conf] action delete (up to date)
  * file[/etc/apache2/sites-available/default.conf] action delete (up to date)
  * link[/etc/apache2/sites-enabled/000-default] action delete (skipped due to not_if)
  * file[/etc/apache2/sites-available/000-default] action delete (skipped due to not_if)
  * link[/etc/apache2/sites-enabled/000-default.conf] action delete
    - delete link at /etc/apache2/sites-enabled/000-default.conf
  * file[/etc/apache2/sites-available/000-default.conf] action delete
    - delete file /etc/apache2/sites-available/000-default.conf
<snip>
* template[/etc/apache2/sites-available/000-default.conf] action create
    - create new file /etc/apache2/sites-available/000-default.conf
    - update content in file /etc/apache2/sites-available/000-default.conf from none to f55856
    --- /etc/apache2/sites-available/000-default.conf	2015-03-03 16:52:20.000000000 +0000
    +++ /tmp/chef-rendered-template20150303-29773-5fs1oj	2015-03-03 16:52:20.000000000 +0000
```

I think maybe amending the not_if guards a bit further to append ".conf" after the default site name to avoid this?
`not_if  { site == "#{node['apache']['default_site_name']}.conf" }`

svanzoest:
Hmm.. unfortunately, it is more complicated than that as you can see that each platform has a slightly different version of the filename for the default site. I think ultimately it is better to not use the default_site at all and only have the option to remove it, if it is provided by the package. If you want to create a default site, you would just use the normal apache_webapp approach streamlining the code significantly.

The biggest reason, I think you are seeing some of these issues is because most people leave default_site_enabled to false. We may want to consider changing the guard to test for that, but I would need to get my head back into the code before I am certain that that is the right approach.

gsreynolds:
Indeed I am encountering this because I've had reasons to set default_site_enabled to true which may well be unusual, but as it stands the cookbook has support for a default site which is broken by a continuous delete and create on each run. 

The default recipe will create a web_app called default or 000-default depending on platform if default_site_enabled is true, which will be output with the filename "default.conf" or "000-default.conf". On the next converge, the default recipe will delete all sites named "default default.conf 000-default 000-default.conf", and then create it again.

I'd agree that adding an extra test to the guards is probably the best way forward e.g.
```
not_if { site == "#{node['apache']['default_site_name']}.conf" && node['apache']['default_site_enabled'] }
```
so the logic of the guard always deletes if default_site_enabled is false, but if it's true only deletes if it doesn't match the "#{node['apache']['default_site_name']}.conf" . Have pushed a further commit with this change which passes ChefSpec.

svanzoest:
Thanks for this, we will see how we deal with default_site going forward, but it definitely should function currently.

gsreynolds:
:+1: thanks!

PR #324, opened 2015-03-06
Allow for custom ssl port and use node.default rather than node.set

tokenrain:
* Enable the user to set the default ssl port to something other than 443 if they want.

* Use node.default rather than node.set for adding ssl port as it should be the same precedence as the other ports. No need for a normal setting here

Please let me know what you think about merging these changes. thanks!

### DIFF
--- a/attributes/mod_ssl.rb
+++ b/attributes/mod_ssl.rb
@@ -18,6 +18,7 @@
 # limitations under the License.
 #
 
+default['apache']['mod_ssl']['port'] = 443
 default['apache']['mod_ssl']['protocol'] = 'All -SSLv2 -SSLv3'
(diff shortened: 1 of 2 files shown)


svanzoest:
Hi! Thanks for using the cookbook. 

I think ultimately the ports logic will go away and we will rely entirely on `apache.listen_ports` and have you supply your own vhost templates. I do not mind accepting something like this for now though. We will want to make sure the tests check for this as well. I would be somewhat tempted to replace that line that adds 443 entirely. Essentially ask the user to specify the ports in their application/wrapper cookbooks.

tokenrain:
I agree with the sentiment. I am in fact going to use a wrapper cookbook to override 443 once this gets merged and bundled. I also know it is hard if not impossible to break compatibility and this is an old and well used cookbook. This seems like the "best" solution for a fix in the short term. Let me know if you need anything else from me

PR #332, opened 2015-03-17
SSLStrictSNIVHostCheck breaks RHEL 5

scottymarshall:
Putting conditional around SSLStrictSNIVHostCheck so it does not appear if it is not turned on. It is turned off by default and if it appears in apache before 2.2.12, it will break the server.

### DIFF
--- a/templates/default/mods/ssl.conf.erb
+++ b/templates/default/mods/ssl.conf.erb
@@ -76,7 +76,7 @@
   #   secure renegotiation protocol. Default: Off
   SSLInsecureRenegotiation <%= node['apache']['mod_ssl']['insecure_renegotiation'] %>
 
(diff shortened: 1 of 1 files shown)


scottymarshall:
My apologies for the bad code.  I have since corrected it.   My goal is for the parameter to not show up if it is set to "Off".  This does not change much because "Off" is the defaut for this parameter, and it also allows backwards compatibility for versions before 2.2.15.

drpebcak:
I don't see any harm in adding this. The default value for apache has this as off as well, so not explicitly saying `SSLStrictSNIVHostCheck Off` won't change any settings.

@svanzoest any issues with this?

svanzoest:
Nope. All good here.

Issue #302, opened 2014-12-22
New attribute needed: service_name


cevou:
Currently the package name attribute is used to install the package and to identify the service. On amazon linux it is possible to install apache 2.4 using the package httpd24. The service is still called httpd.
In this constellation the cookbook fails, because the service httpd24 cannot be found.

Issue #160, opened 2014-07-09
Support service[apache2] graceful instead of reload in web_app definition


runwaldarshu:
In general it will be nice to have graceful option in apache2 service provided by this cookbook which can be used by consumers.

Specific issue seen:
While doing some testing I have found that using service reload have some downtime on RHEL 6.5 httpd service. Whenever there are changes in configuration using web_app definition we would like it to call service graceful instead of reload so that we have no server downtime.

If it is not possible to add graceful option instead of reload in web_app https://github.com/onehealth-cookbooks/apache2/blob/master/definitions/web_app.rb#L40 at least provide a way to configure the action so that user can skip this action by setting it to none and then handle the reload/graceful on their end.

In general it will be nice to use graceful instead of reload if possible. Or at the least provide a way to ovrride or configure that option by consumer cookbooks if needed

bbaja42:
Sorry for commenting on old merged issue:

I noticed that reload command was replaced with graceful.

Since graceful actually performs the restart, then what is the purpose of having both graceful and restart command, why not do graceful always?

drpebcak:
@bbaja42 depending on what platform you are referring to, there is a difference between restart and graceful - basically, [this](https://teckadmin.wordpress.com/2013/10/23/difference-between-graceful-restart-and-normal-restart/).

svanzoest:
This the section in the docs on the difference between restart and graceful.
http://httpd.apache.org/docs/current/stopping.html


On May 12, 2017 09:19, "Taylor Price" <notifications@github.com> wrote:

@bbaja42 <https://github.com/bbaja42> depending on what platform you are
referring to, there is a difference between restart and graceful -
basically, this
<https://teckadmin.wordpress.com/2013/10/23/difference-between-graceful-restart-and-normal-restart/>
.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/sous-chefs/apache2/issues/160#issuecomment-301121566>,
or mute the thread

bbaja42:
Cool, thnx

PR #335, opened 2015-03-31
Use systemd on RHEL/CentOS 7

joerocklin:
On a minimal CentOS 7 system using just systemd (without initscripts installed) the service setup fails because `/sbin/service` isn't installed. The systemd service definitions on CentOS 7 correctly use the `graceful` command for the reload process.


svanzoest:
This looks good to me, can you update the serverspec and chefspec to match?

drpebcak:
This will break Amazon Linux support #302 

Amazon Linux does not use systemd, but it does return `rhel` for `platform_family`

svanzoest:
@drpebcak yes leaving that to chef is better rather than hard coding it.

joerocklin:
Does RHEL/etc < 7 or Amazon need the reload command defined this way external to the package-provided init scripts? Should all of the case statements just be removed?

drpebcak:
I believe the `Chef::Provider::Service::Init::Redhat` class still looks for /sbin/service... not 100% sure though.

svanzoest:
@joerocklin this is because redhat (and fedora) do not use `graceful`, but rather use `restart`. See #160 among others. It seems like they fixed it in RHEL7, which means we can guard it and only use it on lower versions. 

We can not decide upon a server provider and just leave that decision to chef itself. It is much better at that and it is important for that to be consistent across cookbooks.

svanzoest:
where are with this one @drpebcak ?

drpebcak:
Not sure what we wanted to do for this. We could say:

```
if node['platform_version'].to_f < 7.0 || node['platform'] == 'amazon' 
  reload_command '/sbin/service httpd graceful'
else
  provider Chef::Provider::Service::Systemd
end
```

svanzoest:
@drpebcak I wouldn't even specify the provider, just let chef handle that.

```ruby
if node['platform_version'].to_f < 7.0 || node['platform'] == 'amazon' 
  reload_command '/sbin/service httpd graceful'
end
```

drpebcak:
@svanzoest would that work? I thought chef only recognizes one provider per platform (https://github.com/chef/chef/issues/1717). I think we would still have to override it if we wanted to use systemd.

drpebcak:
I'm not sure if [this](https://github.com/chef/chef/pull/1596) was implemented yet either, but this could also potentially fix this problem.

svanzoest:
@drpebcak It was added in chef 12 and chef-client 3.8.2 (via https://github.com/opscode-cookbooks/chef-client/pull/190)

drpebcak:
Do you think we could let newer chef releases pick the provider automatically, but hard code it in some places for backwards compatibility? 

I think that'd be the best way to go.

svanzoest:
I guess we can but I do not know what we would hardcode and what it would buy us. This really not a cookbook issue as each cookbook would have the same issue.

PR #338, opened 2015-04-02
Apache2.4 on Amazon linux puts magic in a different place

jorhett:
This fixes magic location on Amazon Linux.

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -101,8 +101,10 @@ when 'redhat', 'centos', 'scientific', 'fedora', 'amazon', 'oracle'
   default['apache']['cache_dir']   = '/var/cache/httpd'
   default['apache']['run_dir']     = '/var/run/httpd'
   default['apache']['lock_dir']    = '/var/run/httpd'
+  default['apache']['magic_file_dir'] = node['apache']['conf_dir']
   if node['platform'] == 'amazon' && node['apache']['version'] == '2.4'
     default['apache']['devel_package'] = 'httpd24-devel'
+    default['apache']['magic_file_dir'] = '/etc/httpd'
   end
   if node['platform_version'].to_f >= 6
     default['apache']['pid_file'] = '/var/run/httpd/httpd.pid'
--- a/templates/default/mods/mime_magic.conf.erb
+++ b/templates/default/mods/mime_magic.conf.erb
@@ -1,3 +1,3 @@
 <IfModule mod_mime_magic.c>
-  MIMEMagicFile <%= node['apache']['dir'] %>/magic
+  MIMEMagicFile <%= node['magic_file_dir'] %>/magic
 </IfModule>


jorhett:
Curious, why hasn't this been merged?

drpebcak:
Sorry, hadn't had a chance to look at it. Doesn't this break things for other platforms like Ubuntu/Debian?

jorhett:
It's wrapped inside an if statement for Amazon Linux.

drpebcak:
What will happen if you run this on Ubuntu? `MIMEMagicFile <%= node['magic_file_dir'] %>/magic`

jorhett:
I'm really not trying to be rude, but why all the back and forth on a 3 line patch? Is it really unclear how it works?

drpebcak:
Let me put it this way - 
```
case node['platform']
when 'redhat', 'centos', 'scientific', 'fedora', 'amazon', 'oracle'
  default['apache']['magic_file_dir'] = node['apache']['conf_dir']
  if node['platform'] == 'amazon' && node['apache']['version'] == '2.4'
    default['apache']['magic_file_dir'] = '/etc/httpd'
when 'suse', 'opensuse'
  ????
when 'debian', 'ubuntu'
  ????
when 'arch'
  ????
when 'freebsd'
  ????
else
  ????
end
```

What do you expect `MIMEMagicFile <%= node['magic_file_dir'] %>/magic` to resolve to on those other platforms?

It doesn't resolve the same for other platforms because you told it to use a different attribute that you didn't set for the other platforms.

jorhett:
Did you read line 104?  It sets it once for any platform which doesn't override it.

drpebcak:
Please look over the attributes file a little more carefully. Line 104 is within a case statement. While it sets the attribute for a lot of platforms, it doesn't do anything for Ubuntu, Debian, FreeBSD, etc.

svanzoest:
I don't mind pulling this in, but we will need to set `apache.magic_file_dir` on all platforms and add it to the chefspec and serverspec tests.

svanzoest:
The more I look at this it does not actually change the location of the magic file for amazon, it actually moves it for all the other rhel platforms? Is that the intent? Obviously, we want everything to point to the correct place, but this seems to break more than it fixes. 

Feel free to open up a new PR with a better explanation as to what the intent is and we can go from there.

jorhett:
I'm not sure what you are reading. There is a single IF that only adjusts the location if it's amazon linux and apache 2.4. It works properly on all three of RHEL6, RHEL7, and Amazon linux 2013.09 -> 2015.09

svanzoest:
What I am reading is broken code that references `magic_file_dir` which does not exist anywhere and setting `apache.magic_file_dir` only on `rhel` platform_family, but no where else.

On `redhat`, `centos`, `scientific`, `fedora`, `amazon`, `oracle`, the attempt here suggests the resulting `MIMEMagicFile` directive would be `MIMEMagicFile /etc/httpd/conf/magic` which used to be `MIMEMagicFile /etc/httpd/magic` and on `amazon` with `apache.version` set to 2.4, you keep it at `MIMEMagicFile /etc/httpd/magic`

On all other platforms, the code does not define `apache.magic_file_dir` nor `magic_file_dir`, so it breaks the chef run entirely.

That seems to not match any of the intent your are describing, so I am at a loss as to how to move forward. I don't really have time to make educated guesses and then write the tests for them.

Please consider submitting a new PR with chefspec tests for this and make sure they pass on all platforms.

PR #339, opened 2015-04-02
Handle alternate shared object filename for Amazon Linux php55/php56

jorhett:
This is necessary when using Amazon Linux php55 and php56 packages.


svanzoest:
Thanks! 

Currently this causes the chefspec (and possibly serverspec) tests to fail (see Travis CI build failure). Can we update the fixtures by defining `apache.mod_php5.so_filename` for each supported platform in `test/fixtures/platforms` and updating the tests to match?

svanzoest:
This seems to be handled by @drpebcak commits.

PR #340, opened 2015-04-02
Ensure we grab Amazon's mod24_ packages.

jorhett:
Grab Amazon's renamed packages when using 2.4 on amazon linux

### DIFF
--- a/recipes/mod_dav_svn.rb
+++ b/recipes/mod_dav_svn.rb
@@ -22,7 +22,11 @@ include_recipe 'apache2::mod_dav'
 package 'libapache2-svn' do
   case node['platform_family']
   when 'rhel', 'fedora', 'suse'
-    package_name 'mod_dav_svn'
+    if node['platform'] == 'amazon' && node['apache']['version'] == '2.4'
+      package_name 'mod24_dav_svn'
+    else 
+      package_name 'mod_dav_svn'
+    end 
   else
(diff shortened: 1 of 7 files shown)


svanzoest:
We should try to abstract this at a higher level so we do not need to change every single recipe. Maybe we need to create a new attribute that we can conditionally set.

svanzoest:
Maybe something like `node['apache']['mod_prefix'] = 'mod24_'`

jorhett:
I was thinking about that, but there's two things we need to track:

1. module prefix
2. whether or not to install the package -- as it seems it's not necessary except on Amazon?
3. the mod prefix would be rhel family only

jorhett:
I dunno, look at recipes/mod_perl.rb and I think you'll see that the prefix idea will be tricky

jorhett:
Okay, I just spent several hours making this work. Take a look at https://github.com/svanzoest-cookbooks/apache2/pull/341 where I've moved much of the module config into attributes, and given you a default recipe idea which will allow you to remove many of the mod_ recipes.

This is fairly major work, but it's both simple and powerful. There's some followup work to reduce many of the mod_ recipes into attributes, but I've done the brunt of the work for you.

svanzoest:
Thanks for this! It is much appreciated.

svanzoest:
I am going to close this in favor of #341

PR #341, opened 2015-04-02
Major rewrite to put many module configs into attributes

jorhett:
This is a major rewrite to put many module configs into attributes.

Furthermore this consistent logic made it easy to create a "default recipe" for modules which don't need special hacks. Many of the mod_ recipes can be removed after shifting some parts of them into attributes.

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -320,23 +320,3 @@ default['apache']['proxy']['require']    = 'all denied'
 default['apache']['proxy']['order']      = 'deny,allow'
 default['apache']['proxy']['deny_from']  = 'all'
 default['apache']['proxy']['allow_from'] = 'none'
-
-# Default modules to enable via include_recipe
-default['apache']['default_modules'] = %w(
-  status alias auth_basic authn_core authn_file authz_core authz_groupfile
-  authz_host authz_user autoindex dir env mime negotiation setenvif
-)
-
-%w(log_config logio).each do |log_mod|
-  default['apache']['default_modules'] << log_mod if %w(rhel fedora suse arch freebsd).include?(node['platform_family'])
-end
-
-if node['apache']['version'] == '2.4'
-  %w(unixd).each do |unix_mod|
-    default['apache']['default_modules'] << unix_mod if %w(rhel fedora suse arch freebsd).include?(node['platform_family'])
-  end
-
-  unless node['platform'] == 'amazon'
-    default['apache']['default_modules'] << 'systemd' if %w(rhel fedora).include?(node['platform_family'])
-  end
-end
--- a/attributes/modules.rb
+++ b/attributes/modules.rb
@@ -1,0 +1,200 @@
+#
+# Cookbook Name:: apache2
+# Attributes:: modules
+#
+# Copyright 2008-2013, Chef Software, Inc.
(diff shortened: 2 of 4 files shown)


jorhett:
Hey, anything I can do to help move this along?

svanzoest:
Hi @jorhett 
I am looking this over, but having this PR pass the built would be a good step.

I wonder if we can shorten that `has_recipe` hash, as that seems prone for typos and easy to forget adding it to that list when a new module gets added.

jorhett:
Okay, so I'm happy to keep developing towards a major version cut with you, but the hack/slash we have to do to work around the lack of amazon support in apache2 is frustrating, and 5-7x the size of these patches.

How can we move forward to fix the immediate problems with amazon support and get those changes into the supermarket?

jorhett:
"I wonder if we can shorten that has_recipe hash, as that seems prone for typos and easy to forget adding it to that list when a new module gets added."

I can't parse this sentence into an actionable statement. What do you think needs shortened? The name of the hash? The entries? ...?

svanzoest:
I agree. It would be great if we could just have Amazon Linux be more consistent, so we do not need to put all these conditionals in.

>> "I wonder if we can shorten that has_recipe hash, as that seems prone for typos and easy to forget 
>> adding it to that list when a new module gets added."

> I can't parse this sentence into an actionable statement. What do you think needs shortened? The 
> name of the hash? The entries? ...?

I was more thinking that we should see if we could it in a way that made it easier to review at a glance and therefore make it less likely that we would forget to update it. There can be many moving parts in a cookbook like this, and it is important to focus on consistency and ensuring we do not break all the platforms we support. 

In the chefspec/serverspec tests we have arrays `modules_with_config` and `modules_without_config` which is another way to approach this. That is error prone in slightly different ways, such as adding a module to both, etc. The goal is to avoid a maintenance headache and minimize the chance that a typo could get through in a release. I would rather have it fail hard during chefspec then for it to slip by.

jorhett:
Amazon Linux is quite consistent. You can install any of PHP 5.3, 5.5, or 5.6. The prefix changes with each version.

php-      = php 5.3
php55-  = php 5.5
php56-  = php 5.6

This kind of prefixing ensures consistency and lack of surprise. Many/most apps fail when changing PHP versions without warning. Having php- suddenly install php 5.6 would definitely cause major problems.

jorhett:
As for "easier to review at a glance" -- still seems to be subjective, and I'm not a mind reader.  I consider my Pull requests to be suggestions. Trust me that I often refactor Pulls I get from others to my own desires, I really don't care if you do that to mine. But asking me to figure how what naming style will make you happy... isn't really something you should bounce back to me. Adjust to make yourself happy :D

svanzoest:
No worries. I'll take a look at this when I find some time and get it merged in.

svanzoest:
The more I look over this, I think there are some interesting ideas in here. However, this seems to not have any test updates nor does it seem to address other platforms.

Thanks for your work on this and will take it into account in future improvements such as #414

jorhett:
Frankly, asking me to do a lot of work to make it work the way you want and then ignoring it for A YEAR then closing it because it doesn't meet some guideline you didn't mention... wow.

You just love wasting people's time, don't you?

PR #344, opened 2015-04-13
fix typo in attribute file

leoh0:
default -> node


drpebcak:
Not sure why you closed this, but thanks for pointing it out.

leoh0:
Because I just finished testing in my lab and now I'm sure about it. Thanks.

PR #345, opened 2015-04-15
Optional installation using definitions

stefancarlton:
I'm using this cookbook to install Apache and then configure multiple sites. When doing so, it re-runs the "default" recipe attempting to re-installing everything again.

I've created two new optional values which target this behaviour and allow it to be disabled - they are defaulted to true to continue backwards compatibility.

The service definition is split into a separate recipe to allow for inclusion within the definitions.


svanzoest:
Can you describe your use case a bit more? I am trying to understand what
you are doing that is not solved by web_app.

stefancarlton:
Hi,

I'm installing and configuring Apache outside of the "web_app" definition and then using "web_app" to create the vhost files and manage the server (we have multiple vhosts on one server).

In my use case here, running through web_app runs through the installation of Apache each time - each time it checks all the components and finds them installed.

It's not a drama, just figured it was a minor optimisation.

svanzoest:
What configuration are you doing to apache that is not supported by the cookbook? Do you have a use case and example output that you can share? It would be good to get this resolved.

damacus:
hi @stefancarlton are you still looking to achieve this with the cookbook?

Happy to keep open, or close due to inactivity.

PR #352, opened 2015-05-07
Fix icon directory path on newer RHEL-like platforms that use 2.4

ramereth:
I discovered that the icons were broken on CentOS 7.x and realized that the attribute logic needs to be fixed to use the correct path. This PR should fix this.

Thanks!


svanzoest:
Hi,

Thank you for using the cookbook and providing this contribution.

This looks good. However, the `apache.version` attribute is not dynamicly set, so we can not rely on it to be anything other than '2.4' to get the apache 2.4 behavior. Would you mind adjusting this to check for `apache.version` to be '2.4'? Once that is done I will merge it in.

ramereth:
@svanzoest valid point, fixed! Thanks for the quick turnaround :smiley:

svanzoest:
merged #352 . Thanks!

Issue #351, opened 2015-05-06
[Ubuntu] Default recipe installs conflicting mpm_itk.load file


SIGUSR2:
Despite not supporting mpm_itk, when using apache 2.4 the default `mpm_*.rb` recipes are calling the apache_module definition for mpm_itk by doing: `apache_module('mpm_itk') { enable false }`. This will create the `/etc/apache2/mods-available/mpm_itk.load` file, which has the wrong content and conflicts with future unattended installations of the `apache2-mpm-itk` package (called by another cookbook). From a shell (after manually installing the `apache2-mpm-itk` package and electing to keep the existing file): 
```
root@host# cat /etc/apache2/mods-available/mpm_itk.load
LoadModule mpm_itk_module /usr/lib/apache2/modules/mod_mpm_itk.so

root@host# ls -la /usr/lib/apache2/modules/mod_mpm_itk.so
ls: cannot access /usr/lib/apache2/modules/mod_mpm_itk.so: No such file or directory

root@host# ls -al /usr/lib/apache2/modules/*itk*
-rw-r--r-- 1 root root 19000 Jan 10  2014 /usr/lib/apache2/modules/mpm_itk.so
```
During the chef-client run, when attempting to call `package 'apache2-mpm-itk'` from a different cookbook, after converging apache2::default , the error is: 
```
Setting up libapache2-mpm-itk (2.4.6-01-1) ...

Configuration file '/etc/apache2/mods-available/mpm_itk.load'
 ==> File on system created by you or by a script.
 ==> File also in package provided by package maintainer.
   What would you like to do about it ?  Your options are:
    Y or I  : install the package maintainer's version
    N or O  : keep your currently-installed version
      D     : show the differences between the versions
      Z     : start a shell to examine the situation
 The default action is to keep your current version.
*** mpm_itk.load (Y/I/N/O/D/Z) [default=N] ? D
```
and the exit code is 100. 

Tested using v3.0.1 on ubuntu-14.04-x86_64.

svanzoest:
Hi, 

Thanks for your report and using the cookbook. We appreciate it.

[...]

PR #357, opened 2015-05-31
Fix itk enable and load

brauliobo:
ITK depends on prefork module to work. Also fix module filename and install it from package.

### DIFF
--- a/recipes/mpm_itk.rb
+++ b/recipes/mpm_itk.rb
@@ -18,9 +18,13 @@
 #
 
 apache_module('mpm_event') { enable false }
-apache_module('mpm_prefork') { enable false }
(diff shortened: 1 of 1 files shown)


svanzoest:
Hi @brauliobo as of [v3.0.0](https://github.com/svanzoest-cookbooks/apache2/releases/tag/v3.0.0) and #351 we do not directly support the itk mpm module.

It is suggested to create a separate cookbook that handles the unsupported module. We have some instructions in the [README](https://github.com/svanzoest-cookbooks/apache2/blob/v3.1.0/README.md#otherunsupported-mpm) that basically describes this PR. This also allows you to specify exactly which platforms will be supported by your cookbook.

Issue #362, opened 2015-06-16
.htaccess/.htpasswd regex should be escaped


cmann:
In `templates/default/apache2.conf.erb` there's a `File` directive to disallow access to .htaccess/.htpassword files. The `.` character needs to be escaped on lines 96 and 98 as it's currently treated as a wildcard.

Eg. I use the cookbook to help deploy a mirror of the Ubuntu archives and the `shtool` package cannot be downloaded without manually modifying the configuration as it matches the existing expression.

Fix is in pull request #363 

It also looks like the use of `node['apache']['access_file_name']` as both a string (for AccessFileName) and a regex (in file_name_prefix) could cause similar bugs, if I'm not mistaken. If any of the first three characters of the attribute have special meaning in a regex they'd need to be escaped first.

svanzoest:
:+1: thanks. I will merge shortly.

PR #363, opened 2015-06-16
Escape '.' in regex for .htaccess/.htpasswd files

cmann:
Fix for issue #362 

The default suite passes except for the mod_fastcgi tests, which sounds like it's to be expected. Converge fails on default-opensuse-131 due to failing to install mod_ssl but I also get the same result on master.


cmann:
Ah, looks like I need to update `spec/default_spec.rb` too. Working on it.

cmann:
Looks like that did the trick, should be good to go.

PR #365, opened 2015-06-22
force log directory creation to be recursive

joshoohaah:
Right now its not recursive and causes errors if the parent directories aren't created

### DIFF
--- a/recipes/default.rb
+++ b/recipes/default.rb
@@ -50,6 +50,7 @@ end
 
 directory node['apache']['log_dir'] do
   mode '0755'
+  recursive true
 end
 
 # perl is needed for the a2* scripts


drpebcak:
What platform were you seeing problems on?

joshoohaah:
Centos 6.6 .  And Oracle. I would change the folder to something different 
like "/logs/tmp/httpd" and would get errors that the other parent folders 
don't exist.

Sent with AquaMail for Android
http://www.aqua-mail.com

drpebcak:
Aah, ok that makes more sense now. I'll go ahead and merge this.

drpebcak:
Thanks for the pull request!

PR #368, opened 2015-06-30
Change the service creation to use the service_name attribute throughout

rocpatel:
The service_name attribute was only used to define the service.  However,
the restart_command and reload_command for rhel had a hardcoded service_name
in it.  So, for example Software Collections has an apache2.4 rpm package for
redhat 6 that installs the httpd service under a different name 'httpd24-httpd'.
The recipe would fail at service creation even though I overwrote the service_name
attribute.


recipes/default.rb
```diff
@@ -203,16 +203,18 @@
...
   when 'rhel'
-    restart_command '/sbin/service httpd restart && sleep 1' if node['apache']['version'] == '2.2'
-    reload_command '/sbin/service httpd graceful'
+    restart_command "/sbin/service #{apache_service_name} restart && sleep 1" if node['apache']['version'] == '2.2'
+    reload_command "/sbin/service #{apache_service_name} graceful"
   when 'debian'
     provider Chef::Provider::Service::Debian
   when 'arch'
-    service_name 'httpd'
+    service_name apache_service_name
```
svanzoest: This will break `arch` as now it would look for `apache` and not `httpd`. See https://github.com/svanzoest-cookbooks/apache2/blob/master/attributes/default.rb#L181

rocpatel: i could add an attribute under arch which sets node['apache2']['service_name'] = 'httpd'

svanzoest:
Thanks!

PR #371, opened 2015-07-02
Use available attributes in templates

crepetl:
I suggest to use the available attributes inside templates.

### DIFF
--- a/templates/default/web_app.conf.erb
+++ b/templates/default/web_app.conf.erb
@@ -26,10 +26,11 @@
 
   <% if node['apache']['version'] == '2.4' -%>
     Require local
+    Require ip <%=node['apache']['status_allow_list']%>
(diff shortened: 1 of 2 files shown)


templates/default/web_app.conf.erb
```diff
@@ -39,7 +40,7 @@
   LogLevel info rewrite:trace1
   <%- else -%>
   LogLevel info
-  RewriteLog <%= node['apache']['log_dir'] %>/<%= @application_name %>-rewrite.log
+  RewriteLog <%= node['apache']['log_dir'] %>/<%= @params[:name] %>-rewrite.log
```
svanzoest: we will not want to change this, as that is a `web_app` definition parameter that defaults to `@params[:name]`

PR #372, opened 2015-07-20
Allow to use combined attribute for ports.conf

jrosenboom:
Instead of requiring to use the same set of ports on all configured addresses, we
allow to configure an explicit list of address:port pairs to listen on.

This will make it possible have setups listening on e.g.

  10.10.0.1:80
  10.10.0.2:443

without also listening on 10.10.0.1:443 or 10.10.0.2:80.

I know this is still lacking tests, I just would like to get some feedback first on whether you would consider this patch for inclusion. We are using it locally for installing apache2 as proxy in front of our OpenStack setup.

### DIFF
--- a/spec/default_spec.rb
+++ b/spec/default_spec.rb
@@ -25,14 +25,7 @@ describe 'apache2::default' do
           end
 
           it "installs package #{property[:apache][:package]}" do
-            case platform
-            when 'freebsd'
-              expect(chef_run).to install_freebsd_package(property[:apache][:package])
-            when 'amazon', 'fedora', 'redhat', 'centos'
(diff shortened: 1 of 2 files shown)


jrosenboom:
The errors in the unit test seem not related to my patch, I see them also when testing the upstream master locally. I also have a patch for that, https://github.com/x-ion-de/apache2/commit/4b7adde822d68b3c165028f4be3ec784e9839da5, I'm just wondering why the errors weren't seen in the CI for the last upstream commit.

jrosenboom:
Somehow I managed to push the wrong branch here, will close for now and reopen once the CI issues are cleaned up.

PR #374, opened 2015-07-21
Make metadata.rb compatible with versions < 12.

delianides:
Resolves error when generating metadata with versions < 12.


martinb3:
:+1: seems reasonable to me.

PR #378, opened 2015-07-24
change listen attributes to be hash of addresses and ports to listen to

b-dean:
This should fix the issue that was going on w/ PR #372
The attributes `node['apache']['listen']` will be a hash indexed by address where the values are ports to listen to.

For example:

```json
{
  "apache": {
    "listen": {
      "*": ["80", "443"],
      "127.0.0.1": [ "12345" ]
    }
  }
}
```

would create this ports.conf:

```
Listen *:80
Listen *:443
Listen 127.0.0.1:12345
```

Which was impossible with the older listen_ports and listen_addresses attributes.


b-dean:
FWIW, the travis failures were failing before my PR.

martinb3:
I would love to see this made backwards-compatible, checking for the old attribute if it exists & using the older behavior, and emitting a warning that the user should upgrade.

b-dean:
I believe the changes I made to the `ports.conf.erb` makes it still use the old attributes the same way they worked before.

It doesn't give a message about them being deprecated or anything like that though.

b-dean:
Is there some preferred place to have it give a warning about deprecated attributes? I'm guessing that the right place would not be inside the `ports.conf.erb` while it's running the template.

b-dean:
I had been doing the conversion of `node['apache']['listen_addresses']` and `node['apache']['listen_ports']` to the new `node['apache']['listen']` directly in the `ports.conf.erb` template. I moved it out to a helper method where it's easier to see whats going on and we can have a warning message about the deprecated attributes.

b-dean:
Realized I needed to use a `Chef::Mixin::DeepMerge.deep_merge` in my helper method. Also rebased to the latest master.

b-dean:
Build failing because of berkshelf nonsense. @svanzoest, @drpebcak, looks like it might be time to go to berkshelf 4.

drpebcak:
Looks good to me! Thanks!

We'll have to look into the berkshelf thing...

jrosenboom:
Sorry for jumping in late here, but somehow I didn't notice the updates to my original PR that triggered this work. 

Using a hash for the adresses makes using the listen attribute rather complicated, so I'm wondering whether this could get amended to using a flat array of "addr:port" combinations.

The use case I'm having is that a recipe wants httpd to listen on addr:port in addition to whatever some other recipes may have configured. Now instead of being able to simply set `listen += ["addr:port"]`, the code will first have to check whether the `listen` hash already has a key with value `addr` and depending on that either create a new key-value pair or just add `port` to the existing list.

I would also vote to get rid of setting `listen = {'*' => '80']` as a default, since overriding this default turns out to be pretty complicated if one doesn't know whether it might have been set by a cookbook or not. I'll come up with a new PR for that, but would like to know your ideas regarding this anyway.

b-dean:
@jrosenboom that seems fine to me. My original need here was just to be able to have it listen on more than every address on every port. An array of `addr:port` should work too.

The only downside I can think of is that arrays don't merge across node attribute levels the same way hashes do. So the apache2 cookbook couldn't have an attributes file with

```ruby
default["apache"]["listen"] = [ "*:80" ]
```

and then later someone in some `foo::bar` recipe does: 
```ruby
node.normal["apache"]["listen"] += [ "127.0.0.1:1234" ]
```

because then the normal level wouldn't have the `"*:80"` so it wouldn't listen there anymore. Arrays are super obnoxious in Chef. So ... maybe it wouldn't work so well.

PR #379, opened 2015-07-30
Mod_cloudflare yum support

rettigs:
This adds yum support to the mod_cloudflare recipe, and also creates Chefspec tests for it.  Since the apt and yum cookbooks are not specified as dependencies in the main metadata.rb, we're getting the tests to pass by using the apache2_test cookbook.  Whether the apt and yum cookbooks should be added to the main metadata.rb as dependencies is up to you.

### DIFF
--- a/recipes/mod_cloudflare.rb
+++ b/recipes/mod_cloudflare.rb
@@ -17,14 +17,26 @@
 # limitations under the License.
 #
 
-apt_repository 'cloudflare' do
-  uri 'http://pkg.cloudflare.com'
-  distribution node['lsb']['codename']
-  components ['main']
-  key 'http://pkg.cloudflare.com/pubkey.gpg'
-  action :add
-end
-
-package 'libapache2-mod-cloudflare' do
-  notifies :restart, 'service[apache2]'
+case node['platform_family']
+when 'debian'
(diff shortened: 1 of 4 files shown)


svanzoest:
Thanks for this. Since `mod_cloudflare` is not something provided by the ASF, we should really pull this out and make it its own separate cookbook. Similar to https://github.com/drpebcak/mod_pagespeed_cookbook.

We generally do not include the dependencies for `apt` and `yum`, `zypper`, etc. as those are platform specific and we do not want to be in a situation where when those are not needed, you are still held back from deploying the cookbook. We talk about that a bit at https://github.com/svanzoest-cookbooks/apache2#notes-for-rhel-family but we should expand on that in more general terms.

martinb3:
It looks like there's also a [mod_pagespeed recipe](https://github.com/svanzoest-cookbooks/apache2/blob/master/recipes/mod_pagespeed.rb) in the apache2 cookbook as well. Should we take that out?

svanzoest:
Yes once the other cookbook is released.

drpebcak:
There is actually an initial release for the mod_pagespeed cookbook: https://supermarket.chef.io/cookbooks/mod_pagespeed

Please check it out and report any problems!

PR #382, opened 2015-09-14
Fix for NameError in mod_proxy_html.rb

vaskur:
Without this change, using this recipe results in:

NameError
---------
No resource, method, or local variable named `platform_family' for `Chef::Recipe "mod_proxy_html"'


drpebcak:
Thanks for the fix.

PR #384, opened 2015-09-15
Fix names for a2enconf and a2disconf

adrianmoisey:


drpebcak:
Thanks for the correction!

PR #389, opened 2015-10-10
Fixed attributes for SuSE

franznemeth:
I have made some fixed for the cookbook when run on SLES:
  - The node['apache']['mod_php5']['so_filename'] attribute doesnt work with SLES, because the PHP module is called mod_php.so not libphp.so.   
  - The attribute node['apache2]['version']  does not work with SLES11 right now because it only has apache2.2 and the cookbook installs a httpd.conf for apache2.4. platform? does not support 'opensuse' only 'suse' and the apache 2.4 package is available from 12.1 upwards.

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -65,9 +65,8 @@ default['apache']['version'] =
     node['platform_version'].to_f >= 18 ? '2.4' : '2.2'
   when 'suse'
     case node['platform']
-    when 'opensuse'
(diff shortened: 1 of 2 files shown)


franznemeth:
FYI: The travis build failed on something else than my changes!

drpebcak:
I'll just cherry-pick the important commit off of this

PR #393, opened 2015-10-28
mention mod_actions in README

dastergon:
Although mod_actions is supported, it was not mentioned in README.


drpebcak:
Thanks for the update!

PR #394, opened 2015-11-06
Properly set ServerLimit on 2.4

nickjwebb:
ServerLimit was not set for apache 2.4. I assume that was an omission, but perhaps not. We have hosts that need > 256 `MaxRequestWorkers`, which is the apache compile time default for `ServerLimit`.


drpebcak:
Thanks!

PR #397, opened 2015-11-17
Add default server name to the apache2.conf template

shanman190:
In my configuration, I am running with `apache httpd 2.2` on RHEL 6.6. In this configuration restarting the httpd service takes upward of 30 seconds due to the server name not being defined in the httpd.conf/apache2.conf file ultimately falling back on localhost for the default. However, due to this extended time taken trying to restart the httpd service the chef installation times out making it impossible to continue provisioning of the chef run. 

This PR adds the `ServerName` directive to the apache2.conf template thus allowing the httpd service to not spend so much time attempting to find the best suited `ServerName`. I've made it configurable based on the `node['apache']['server_name']` attribute and if that is not provided taking the node name instead.

### DIFF
--- a/templates/default/apache2.conf.erb
+++ b/templates/default/apache2.conf.erb
@@ -3,6 +3,7 @@
 #
 # Based on the Ubuntu apache2.conf
 
+ServerName "<%= node['apache']['server_name'] || node.name %>"
 ServerRoot "<%= node['apache']['dir'] %>"
 
 #


svanzoest:
Are you setting the fqdn hostname one the box? This should be resolved relatively quickly.

shanman190:
In my case I'm not really even using the field, but yes I'm setting it to the fqdn of the server. Havjng the ServerName set in this scenario allowed me to get the service restart times down to a reasonable, expected level.

svanzoest:
Have you tried actually setting the `fqdn` hostname on the actual operating system level? That should resolve this and likely lots of other things on your machine. Explicitly defining `ServerName` somewhat hides the problem.

svanzoest:
If you want to actually set this you can use `apache_config` to drop a template configuration which will set this.

josephholsten:
I would be okay with setting a `ServerName` if it's provided, but I think it's unwise to default to `node.name` with no way to unset it.

svanzoest:
@josephholsten the generally is very little reason to set `ServerName` outside of apaches default use of `node.fqdn`. You really only want to set `ServerName` within a `<VirtualHost>` context.

PR #401, opened 2016-01-12
Remove forced deflate Module

CBeerta:
For any distro but Debian mod_deflate is 'forced' upon the system without a chance to disable it or adapt the configuration.

Moving this to the "default_modules" in the attributes should do the same, but allows me to disable it.


svanzoest:
This solution seems to no longer install `deflate` for debian platform family. I think it would be better to add it to the default_modules array itself.

PR #403, opened 2016-01-15
swat integration tests for apache2 cookbook

melezhik:
Hi  @svanzoest !

I have just created an [integration tests](https://sparrowhub.org/info/swat-apache2-cookbook?v=0.000006) for apache2cookbook using [swat](https://swatpm.org/) -  DSL for rapid web tests development. 

To run them:

```      
$ kitchen test swat
```

### DIFF
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -138,6 +138,11 @@ suites:
   attributes:
     machine_fqdn: working.computers.biz
     machine_fqdn_as_hostname: true
+    sparrow:
+      plugin:
+        list: [
+            'swat-apache2-cookbook'
+        ]
   run_list:
   - recipe[fqdn]
   - recipe[apache2_test::setup]
@@ -148,10 +153,4 @@ suites:
     'centos-6.5',
     'debian-7.6'
   ]
-  attributes:
-    sparrow:
-      plugin:
-        list: [
-            'swat-apache2-cookbook'
-        ]
 
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -145,7 +145,8 @@ suites:
   - recipe[apache2_test::basic_web_app]
   - recipe[sparrow]
   includes: [
-    'centos-6.5'
+    'centos-6.5',
+    'debian-7.6'
   ]
   attributes:
     sparrow:
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -143,6 +143,14 @@ suites:
   - recipe[apache2_test::setup]
   - recipe[apache2_test::default]
   - recipe[apache2_test::basic_web_app]
+  - recipe[sparrow]
   includes: [
     'centos-6.5'
   ]
+  attributes:
+    sparrow:
+      plugin:
+        list: [
+            'swat-apache2-cookbook'
+        ]
+
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -134,3 +134,15 @@ suites:
   - recipe[apache2_test::basic_web_app]
   - recipe[apache2_test::broken_conf]
   - recipe[apache2_test::modules]
+- name: swat
+  attributes:
+    machine_fqdn: working.computers.biz
+    machine_fqdn_as_hostname: true
+  run_list:
+  - recipe[fqdn]
+  - recipe[apache2_test::setup]
+  - recipe[apache2_test::default]
+  - recipe[apache2_test::basic_web_app]
+  includes: [
+    'centos-6.5'
+  ]
--- a/Berksfile
+++ b/Berksfile
@@ -7,6 +7,7 @@ group :integration do
   cookbook 'yum', '~> 3.8.2'
   cookbook 'zypper', '~> 0.2.1'
   cookbook 'pacman', '~> 1.1.1'
+  cookbook 'sparrow'  
   cookbook 'fqdn', :git => 'https://github.com/drpebcak/fqdn-cookbook.git'
 end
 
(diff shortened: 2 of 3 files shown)


melezhik:
Hi guys! Any news here?

drpebcak:
Looks like Travis started failing after your two most recent commits.

drpebcak:
Or rather, there is at least one more error since your last two commits.

melezhik:
Hi @drpebcak  my modification do not touch recipe logic at all, please take a look at Travis report

melezhik:
Looks like "upstream" version is not healthy , https://travis-ci.org/svanzoest-cookbooks/apache2/builds/95664672 - all recent travis jobs for this have been failing ...

drpebcak:
I was looking specifically at this: https://travis-ci.org/svanzoest-cookbooks/apache2/builds/103554685#L267

drpebcak:
@svanzoest any thoughts on including another suite of tests?

melezhik:
> I was looking specifically at this: https://travis-ci.org/svanzoest-cookbooks/apache2/builds/103554685#L267

fixed here - 93a8d67ec98b26fccf295aa15ad12b07c2b0a7c4

melezhik:
Hi guys! Any thoughts on possible collaboration?

svanzoest:
I am okay with adding additional test suites as long as they will be maintained. Currently we have chefspec and serverspec that we are maintaining. I do not have the bandwidth to maintain more than that across all of our platforms. I could see us support policyfile type audits/tests to stay aligned with chef corporate, but that would likely mean forgo-ing serverspec in the long run. 

@melezhik what is your intent and goals for collaboration?

melezhik:
Hi ! @svanzoest ! Thanks for feedback.

My intention is to "apply" [swat DSL](https://github.com/melezhik/swat) to test various web applications. There are a lot of examples you could take a look at [SparrowHub](https://sparrowhub.org/). Recently I came up with interesting idea of using SWAT DSL in testing chef cookbook developments. Why?


* Swat tests are very agile and fast to develop - it takes a few time ( literally minutes ) to create a new tests for even complex web application ( http headers, cookies, html forms, named parameters, serial requests, post, gets, heads etc )

* I can't see real alternative for complex integration testing for chef / ansible / <whatever> configuration development , see next  

* Chefspec /  serverpsec are fine but they don't allow you to write complex tests with sending complex  server http requests  and then analyzing  output which. So there is real need for such a tool. Even an [InSpec](https://github.com/chef/inspec) won't cover all the gaps here.

* Swat could be easily integrated and cross platformed as it relies on Perl - mostly pre-installed for many platforms and if even not - [sparrow cookbook](https://supermarket.chef.io/cookbooks/sparrow) cares about it

* Swat tests suites are run via standard bash plugin for test kitchen - a few lines of code - they are really easy to deliver 

* Swat tests are kind of blackbox - so in some way are more objective. They could be developed in parallel with cookbooks. They could be used as specification. (TDD,BDD). They are easy to debug and very pragmatic, as idea behind SWAT - send request and analyze an output . They don't require knowledge of perl to be used. As tests fails you may `kitchen ssh` and just analyze server output gets saved in tetx file, that is very simple.

As for the apache2 cookbook consider a following use cases:

- checking basic/digest, OPEN_ID authentication ( mod_auth , mod_auth_open_id... )
- checking https ( mod_ssl ) - self signed or CA certificates 
- checking virtual hosts via sending dedicated http requests
- checking allow access directives via sending http requests 

( Imagine to implement this with serverspec :smile: )

So in short word I guess our collaboration could be helpful in "both directions". The PR I have just sent is for demonstration purposes only. I'd like to hear if there are real needs for for such test automation beside those one I listed. At least as I can see apache2 cookbook provides many configuration choices for end user , not only installing bare bones apache server, if so I should|could be tested automatically and for many platforms, which could be achievable by SWAT.

svanzoest:
Hi @melezhik,

I agree that those are all important things to do, but they seem to be more useful at a higher level when implementing actual websites, not just the basic deployment of them.

For example, we do not know what virtualhosts someone would want, what SSL settings they plan to use or even what Apache httpd directives someone will have configured.

It makes more sense to have folks use it in those wrapper cookbooks as this level none of the intent you are capturing is known. We just provide the building blocks for them to configure those things easier with sane defaults. We actually plan to remove some of the non-ASF provided modules, so they could be maintained separately and therefore have more focus as we only have bandwidth for so much.

I do not see how to apply swat at this level and would be another thing to get up to speed on, if I do not use it elsewhere. As you say, it would be best to target Web Applications themselves.

melezhik:
Hi @svanzoest , thanks for reply . I do see your considerations. However sometimes difference between providing building blocks for something and building something not that apparent , so if you will however in the future need some tool for web test automation , swat could be one the choices for you. The example I provided just show the way how one could easy embed integration testing into chef cookbook developments . Regards . Alexey

svanzoest:
definitely. It would be good to see how it would fit in compared to [Capybara](http://jnicklas.github.io/capybara/),  [redbot](http://mnot.github.io/redbot/), [har](http://www.softwareishard.com/blog/har-12-spec/), [gauntlt](http://gauntlt.org/), [casperjs](http://casperjs.org/), [watir](http://watir.com/), [rspec-api](http://rspec-api.github.io/) and the like. So much choice around testing that you need to look at which one makes the most sense. It seems like [swat](https://github.com/melezhik/swat) definitely has a place I just do not see me implementing it currently in the `apache2` cookbook.

Thanks again for informing us of swat.

Issue #259, opened 2014-10-09
mod_unique_id cause Apache to fail restart on first coverge


niven01:
Hello,

I've encountered a problem after introducing unique_id where Apache fails to restart.
The work around at this moment is a `killall httpd` and re run chef-client

This can be reproduced in test-kitchen by just including two recipes:  

-
    name: upstream_apache
    run_list:
      - "recipe[apache2]"
      - "recipe[apache2::mod_unique_id]"

The behavior seen is that httpd is stopped but only the master process is killed. There are a bunch of httpd orphans still sitting there listening on 80 and 443

When Apache tries to start it fails as ports are already in use.

Every now and then it will succeed which suggests to me this maybe a timing issue and that httpd start is happening before the stop has completed. 

Error:

```      ================================================================================
           Error executing action `restart` on resource 'service[apache2]'
           
       ================================================================================
           
           Mixlib::ShellOut::ShellCommandFailed
           ------------------------------------
           Expected process to exit with [0], but received '1'
           ---- Begin output of /sbin/service httpd restart ----
           STDOUT: Stopping httpd: [  OK  ]
           Starting httpd: [FAILED]
           STDERR: [Thu Oct 09 11:56:26 2014] [warn] NameVirtualHost *:80 has no VirtualHosts
           (98)Address already in use: make_sock: could not bind to address [::]:80
           (98)Address already in use: make_sock: could not bind to address 0.0.0.0:80
           
       no listening sockets available, shutting down
           Unable to open logs
           ---- End output of /sbin/service httpd restart ----
           Ran /sbin/service httpd restart returned 1
           
           
       Resource Declaration:
           ---------------------
           # In /tmp/kitchen/cache/cookbooks/apache2/recipes/default.rb
           
            24: service 'apache2' do
            25:   service_name node['apache']['package']
            26:   case node['platform_family']
            27:   when 'rhel'
            28:     reload_command '/sbin/service httpd graceful'
            29:   when 'debian'
            30:     provider Chef::Provider::Service::Debian
            31:   when 'arch'
            32:     service_name 'httpd'
            33:   end
            34:   supports [:start, :restart, :reload, :status]
           
           Compiled Resource:
           ------------------
           # Declared in /tmp/kitchen/cache/cookbooks/apache2/recipes/default.rb:24:in `from_file'
           
           service("apache2") do
           
         action [:enable, :start]
             updated true
             supports {:restart=>true, :reload=>true, :status=>true, :start=>true}
[...]

PR #408, opened 2016-02-09
resolve reload problem on rhel based systems using apache2.2

niven01:
Ref: https://github.com/svanzoest-cookbooks/apache2/issues/259

Added `&& sleep 1' if node['apache']['version'] == '2.2'` to RHEL reload


svanzoest:
Resolved in 	09ccbae

PR #409, opened 2016-02-10
Use an array as listen attribute instead of a hash

jrosenboom:
This PR amends the logic introduced in https://github.com/svanzoest-cookbooks/apache2/pull/378 to use an array as attribute instead of a hash, as this will simplify using this cookbook in other recipes.

E.g., if a recipe wants httpd to listen on `addr:port`, it is much easier to
just set

    node.set['apache2']['listen'] += ['addr:port']

instead of first having to check whether the listen hash does have the
key `addr` already and then either add it or update the port list.


jklare:
Any chance we could merge this? The current implementation in master makes it nearly impossible to override the default setting for *:80 and *:443 without using drop on the Mash itself. Additionally the implementation as a Hash leads to chef handling this as a Mash and auto merging all precedences -> meaning you can only override *:80 by setting some other value for the key * (which will always lead to apache listening on * and some port) or dropping it as mentioned above (which is quite ugly).

drpebcak:
Yeah, I kind of like this.

Issue #189, opened 2014-07-29
Use Upstart service Provider on Ubuntu


svanzoest:
See #151 and #150 for some background. 

Related to https://bugs.launchpad.net/ubuntu/+source/apache2/+bug/689454

Chef reverted changing the default service provider to Upstart for >= 13.10 during regression; as this broke many packages without upstart scripts. See opscode/chef#1587

svanzoest:
I tempted to remove the `provider` section in the service block and just let chef-client handle this, however we had #151 which added d42349dca3a24373141375cee49ab044fd468ff8 So, we will need to test if this is still an issue with chef 12 client or determine when this got fixed and require a chef version newer then that bug fix.

drpebcak:
Perhaps we could look into adopting something like [this](https://github.com/poise/poise-service)?

svanzoest:
@drpebcak possibly, but that should be new issue.

drpebcak:
I just mentioned it because it relates. Also with #335.

svanzoest:
I am going to close this, as we have cleaned the provider a lot based on other issues and ultimately, chef-client should manage this for us.

PR #410, opened 2016-02-12
Make sure Upstart is used on Ubuntu 14.04

MrSam:
Add the Chef::Provider::Service::Upstart provider for Ubuntu 14.04


svanzoest:
I do not think we want to hardcode for upstart on any newer version especially because all the newer versions of Ubuntu use systemd. See #189 for some history. You may need to update your version of chef-client to have it use the right provider such as upstart.

Issue #415, opened 2016-03-21
mod_fastcgi URL is unreachable. Cannot run the recipe anymore


jujugrrr:
https://github.com/svanzoest-cookbooks/apache2/blob/master/attributes/mod_fastcgi.rb#L20 doesn't work anymore. It seems www.fastcgi.com is gone

http://webcache.googleusercontent.com/search?q=cache:e0DGxuMJ684J:www.fastcgi.com/+&cd=1&hl=en&ct=clnk&gl=uk

http://www.fastcgi.com

svanzoest:
Do you have an updated URL the cookbook should use? By default it should use the package, so it should not impact people immediately.

matthyatt:
I've been having the same issue. I've only been using chef a few months, but if I read the recipe correctly, when the platform_family is 'rhel' it will always use the source install method and only uses the package on platform_family debian when the install_method is not 'source'. I'm using this cookbook on centos and this recipe attempts to download the source tarball and fails. There's no guard for the remote_file resource so it tries to download it on every converge.

Our current workaround is to self-host the source package.

https://github.com/svanzoest-cookbooks/apache2/blob/master/recipes/mod_fastcgi.rb#L35

jujugrrr:
> Our current workaround is to self-host the source package.

Same here @svanzoest

svanzoest:
It is generally based to leverage package install methods everywhere over dynamic source based downloads as they can be unreliable as seen here.

It looks like the `rhel` platform family does not have a package for `mod_fastcgi` available in the the standard operating system package repositories. It seems that they expect you to use `mod_proxy_fcgi` or `mod_fcgid` instead. What would you consider the best course of action? I have made the default a package install on all platforms in #416, however, there is no way we could have any reliable test coverage as the package is not available by default.

Considering this module is not part of the ASF httpd project itself, I would suggest this gets removed from the `apache2` cookbook entirely and made into its own chef cookbook. Since you all are dependent on this, do you plan to take that on? If so, I would be happy to remove it from the cookbook and direct others your way.

matthyatt:
I would generally agree that since mod_fastcgi isn't available through the ASF httpd source that it shouldn't be included in this cookbook. Installing from a source package is the only way for `rhel` but this package hasn't been updated since 2007 and their site is down so I don't what the future for the mod_fastcgi package is.

As for working with what's in place now, I would think adding a guard to the remote file resource like the bash resource could be a quick fix. 

[...]

PR #417, opened 2016-03-25
Correct fastcgi package for rhel

scassiba:
I found the correct package for the rhel platform family to be mod_fcgid.


svanzoest:
Thanks for looking, however, that would be `apache2::mod_fcgid`. It is not the same, but similar. It should be pretty easy to just switch to that module instead.
See #415 for more on this.

PR #420, opened 2016-04-13
Allow auto-conversion if either of node['apache']['listen_ports'] or …

sdadh01:
…node['apache']['listen_addresses'] are set rather than '&&'. This ensures conversion occurs if only one of the two is set.


svanzoest:
This is related to #409

PR #425, opened 2016-04-27
add node['apache']['mpm_support'] to general settings in README.md

sdadh01:
No added functionality, just adding a line to the docs.

I was bitten by this: I was removing mpm_event using apache_module in a cookbook causing a race condition every chef run as it was installed again by the apache2 cookbook because of defaults on the next run. The attribute was defined in the attributes/default.rb but not documented. Added to documentation just to reduce the small amount of time it takes to resolve any future issues for others looking at the same thing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,6 +178,7 @@ values are noted.
 * `node['apache']['sysconfig_additional_params']` - Additionals variables set in sysconfig file. Default is empty.
 * `node['apache']['default_modules']` - Array of module names. Can take "mod_FOO" or "FOO" as names, where FOO is the apache module, e.g. "`mod_status`" or "`status`".
 * `node['apache']['mpm']` - With apache.version 2.4, specifies what Multi-Processing Module to enable. Defaults to platform default, otherwise it is "prefork"
+* `node['apache']['mpm_support']` - Array of supported MPM modules to install. Default is prefork, worker and event.
 
 The modules listed in `default_modules` will be included as recipes in `recipe[apache::default]`.
 


svanzoest:
I would suggest explicitly define `apache.mpm` to get a specific mpm. That should override whatever the defaults.

sdadh01:
What if a particular installation wants to be able to install two out of
three MPMs? I'm not sure this is really a good use case but
node[:apache][:mpm_support] allows for this. What is happening here is that
a client want's a legacy machine to be controlled by Chef with minimal
changes to the original configuration. They want worker and prefork
avaliable (even though they only use prefork) and not have event installed.
I can do this using mpm_support but it took more time to work out that I
could - hence adding the documentation line.

I'm not sure if we're going for 'if it's possible it should be documented'
or 'only document the things we want people to use'. I favour the former
because if it's in the default attributes file *someone* is going to use it
when they find it (read me ;-)). Just my $0.02...

On Tue, Apr 26, 2016 at 6:03 PM Sander van Zoest <notifications@github.com>
wrote:

> I would suggest explicitly define apache.mpm to get a specific mpm. That
> should override whatever the defaults.
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/svanzoest-cookbooks/apache2/pull/425#issuecomment-214934801>
>
-- 
adh
--
Andrew Hollingsworth
adh@techopsguru.com

Website : http://www.techopsguru.com
Twitter : twitter.com/techopsguru

svanzoest:
Looks like this is related to #236 
Hmm.. I will have to think this through a bit more.

damacus:
@svanzoest any update on this? Should we pull this in or just close this PR?

svanzoest:
@damacus I am fine with pulling this in. I am mostly was concerned with exposing some of the inner workings as user facing documentation. I wouldn't expect anyone to alter `apache.mpm_support` unless they were adding support for additional  DSO MPM modules.

damacus:
@sdadh01 could resolve those merge conflicts so we can pull this in please?


Thanks!

damacus:
Closing due to inactivity.

PR #428, opened 2016-06-03
add variable module support to web_app definition

majuscule:
This patch allows a `web_app` user to specify the modules loaded with the application instead of forcing the hardcoded list. Supports `mod_` and `mod_modname` references, as well as `wrapper-cookbook::module_name`.


svanzoest:
I think we should just remove the modules all together. The only reason they are included is because the default template references directives provided by those modules.

I think the default template should be revisited and made significantly more minimal and rely on individuals to provide a template for the more complex cases.

thoughts?

drpebcak:
I always thought that would be good, but there seems to be an awful lot of resistance to making the default template LESS configurable...

PR #430, opened 2016-06-16
Use plain php package for mod_php5 in Ubuntu 16.04

mattlqx:
Ubuntu 16.04 has dropped the "5" from php packages mean php5 just becomes php now. php7 is now available with number specification.

### DIFF
--- a/recipes/mod_php5.rb
+++ b/recipes/mod_php5.rb
@@ -27,7 +27,11 @@ end
 
 case node['platform_family']
 when 'debian'
-  package 'libapache2-mod-php5'
+  if node['lsb']['release'].to_f >= 16.04
+    package 'libapache2-mod-php'
+  else
+    package 'libapache2-mod-php5'
+  end
 when 'arch'
   package 'php-apache' do
     notifies :run, 'execute[generate-module-list]', :immediately


mattlqx:
Oh, nevermind, libapache2-mod-php is php7 and not 5 like I thought. There should probably be a whole mod_php7 recipe for this.

svanzoest:
thanks for the heads up. It looks like `libapache2-mod-php` maybe a meta package that would provide the latest stable `libapache2-mod-php` version.

JohGiz:
Trying to get mod_php to get going in Ubuntu 16.04 and I get this error.
Using apache2::mod_php5 will try to install libapache2-mod-php2
```
when 'debian'
  package 'libapache2-mod-php5'
```
In Ubuntu 16.04 it would be libapache2-mod-php or libapache2-mod-php7.0 to specify version.

Can't see that this is working or fixed. Or is it?

svanzoest:
Looks like `libapache2-mod-php5` was removed since Ubuntu 16.04 [Xenial](http://packages.ubuntu.com/search?suite=xenial&searchon=names&keywords=libapache%20php).
We should probably make this recipe more generic and remove the version, rather than separating out php4 versus php5 versus php7, etc. and look at making it more generic similar to https://github.com/chef-cookbooks/php/pull/169

svanzoest:
This is resolved in fd478c27. Thanks for the report!

MrMichaelWill:
So in order to support dotdeb php7.0 packages on debian 8.6 (jessie) should I be implementing an apache2::mod_php7.0 ?

PR #432, opened 2016-06-21
Fix wrong regex match on *ht

MrSam:
When using this cookbook all urls with *ht in the path are rejected. 

It generated the following config for 2.4:
<Files ~ "^.ht">
 Require all denied
</Files>

This blocks http://acme.com/xht-blablabla/blabla


svanzoest:
We will want to write some tests for this and look how this relates to the changes in #314

PR #433, opened 2016-06-27
Allow for override of default-site.conf template

burnzy:
These attributes will allow others to override the default site template from their wrapper cookbooks.
The **node['apache']['default_site_template']** attribute will allow users to override the template file, and the **node['apache']['default_site_cookbook']** will allow users to set the cookbook location of where the template file is located.

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -286,6 +286,8 @@ default['apache']['locale'] = 'C'
 default['apache']['sysconfig_additional_params'] = {}
 default['apache']['default_site_enabled'] = false
 default['apache']['default_site_port']    = '80'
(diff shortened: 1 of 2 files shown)


svanzoest:
I am trying to wrap my brain around the full use case around this. 

The default site is usually only used during development to and provides a "It works!" type page. Nothing stops you from putting a `web_app` in your own wrapper cookbook.

burnzy:
@svanzoest Our use case was to utilize the default site for our aws elb healthcheck endpoint - but I suppose a good alternative would be to do like you said and just use another vhost for this purpose. That'll work; thanks for considering this anyway.

Cheers,
Mike

PR #436, opened 2016-07-25
Do not enable alias.conf. Only link alias.load

funzoneq:
mod_alias is enabled by default. It links both the alias.load and alias.conf

Alias.conf contains the following:

```
<IfModule alias_module>
  #
  # Aliases: Add here as many aliases as you need (with no limit). The format is
  # Alias fakename realname
  #
  # Note that if you include a trailing / on fakename then the server will
  # require it to be present in the URL.  So "/icons" isn't aliased in this
  # example, only "/icons/".  If the fakename is slash-terminated, then the
  # realname must also be slash terminated, and if the fakename omits the
  # trailing slash, the realname must also omit it.
  #
  # We include the /icons/ alias for FancyIndexed directory listings.  If
  # you do not use FancyIndexing, you may comment this out.
  #
  Alias /icons/ "/usr/share/apache2/icons/"

  <Directory "/usr/share/apache2/icons">
    Options Indexes MultiViews
    AllowOverride None
    Order allow,deny
    Allow from all
  </Directory>
</IfModule>
```

If you append /icons/ to any domain running this cookbook, you will get a directory listing with icons:

screen shot 2016-07-25 at 13 29 22

Not only is this a useless feature, but it's also a security vulnerability: http://www.acunetix.com/blog/articles/directory-listing-information-disclosure/

### DIFF
--- a/recipes/mod_alias.rb
+++ b/recipes/mod_alias.rb
@@ -18,5 +18,5 @@
 #
 
 apache_module 'alias' do
-  conf true
+  conf false
 end


svanzoest:
Thanks for the report.

I will consider this change, but this really an issue based on the upstream [debian package](
https://packages.debian.org/sid/apache2) and a [bug report](https://bugs.debian.org/cgi-bin/pkgreport.cgi?pkg=apache2) should be filed there as well.

recipes/mod_alias.rb
```diff
@@ -18,5 +18,5 @@
 #
 
 apache_module 'alias' do
-  conf true
+  conf false
```
svanzoest: This specifies that a configure exists and should be installed. If you remove this a config will never be deployed and would need to be deployed manually breaking parity with ASF defaults.

damacus:
Closing due to inactivity.

PR #439, opened 2016-08-05
Make LogFormat configurable

natbraun:


natbraun:
@svanzoest thanks for your feedback on this one.

svanzoest:
I am thinking you may have misunderstood. Those `LogFormat` directives just specify the standard log formats that people would use as a convenience. You would then reference them from [CustomLog](http://httpd.apache.org/docs/current/mod/mod_log_config.html#customlog) or [GlobalLog](http://httpd.apache.org/docs/current/mod/mod_log_config.html#globallog).

You can drop your own [LogFormat](http://httpd.apache.org/docs/current/mod/mod_log_config.html#logformat) directive in a `conf.d` configuration file snippet and then reference it in there or in a `web_app` vhost configuration.

What is the goal of this PR that is not already supported as is?

PR #441, opened 2016-09-01
Fixes for openSUSE / Fedora & general cleanup

tas50:
Fedora and openSUSE were very broken. I've fixed those and I did a bit of cleanup on the testing. There's more to go, but this is a start towards getting everything passing and getting it all in travis.


Berksfile
```diff
@@ -3,10 +3,8 @@ source 'https://supermarket.chef.io'
 metadata
 
 group :integration do
-  cookbook 'apt', '~> 3.0.0'
-  cookbook 'yum', '~> 3.10.0'
-  cookbook 'zypper', '~> 0.2.1'
-  cookbook 'pacman', '~> 1.1.1'
```
svanzoest: was the removal of `zypper` and `pacman` intentional? Those were needed for arch and opensuse.

tas50: arch testing is removed since you can't install omnibus on arch.  zypper isn't actually required. zypper auto refreshes the repos when you install. all that cookbook does is a refresh.

svanzoest: At this point, I do not think it makes sense to remove arch support as it was added. I guess we can reach out to the original contributor and get their feedback. We do want to make sure we are using the latest, it is the same reason apt, yum are there.

svanzoest:
Hey thanks for this. I will look it over in the next few days.

svanzoest:
looks like this broke the chefspec tests?
```
$ bundle exec rspec spec/
bundler: failed to load command: rspec (/home/travis/build/svanzoest-cookbooks/apache2/vendor/bundle/ruby/2.1.0/bin/rspec)
Errno::ENOENT: No such file or directory @ rb_sysopen - /home/travis/build/svanzoest-cookbooks/apache2/spec/support/../../test/fixtures/platforms/debian/7.9.json
```

Issue #414, opened 2016-03-11
convert definitions to custom resources


svanzoest:
Since the release of chef 12.5 we have access to [custom resources](https://docs.chef.io/custom_resources.html) that seem like a more suited method versus using LWRPs for this cookbook #178

stissot:
Unlike Definitions, Custom resources support features like guards (`only_if`, `not_if`)

damacus:
yeah, this is done now. 
They're not complete. But there are now no definitions left on master.

PR #442, opened 2016-09-08
allow to reload web_app immediately via param for definition

jklare:
* Sometimes it's needed to immediately reload a web_app/apache2 to
continue the chef-run, since the webserver is lateron used. This patch
allows to set the reload_speed for the web_app definition (apache2
service) via the parameter :reload_speed. The default behaviour does not
change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -557,6 +557,8 @@ combine the template with enabling a site, see `web_app`.
 
 * `name` - Name of the site.
 * `enable` - Default true, which uses `a2ensite` to enable the site. If false, the site will be disabled with `a2dissite`.
+* `reload_speed` - Defaults to `:delayed` and can be set to `:immediately` to
+  reload apache service directly after web_app definition during the chef-run
 
 web\_app
 --------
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -557,6 +557,8 @@ combine the template with enabling a site, see `web_app`.
 
 * `name` - Name of the site.
 * `enable` - Default true, which uses `a2ensite` to enable the site. If false, the site will be disabled with `a2dissite`.
+* `reload_speed` - Defaults to `:delayed` and can be set to `:immediately` to
```
punkle: chef calls the `speed` field `timer`. Should this be called `reload_timer`

definitions/web_app.rb
```diff
@@ -37,7 +37,7 @@
       :params           => params
     )
     if ::File.exist?("#{node['apache']['dir']}/sites-enabled/#{application_name}.conf")
-      notifies :reload, 'service[apache2]', :delayed
+      notifies :reload, 'service[apache2]', params[:reload_speed]
```
punkle: The web_app is not enabled until the apache_site definition below kicks in. As such, I think this notify would be better placed on apache_site below. However because the apache_site is a definition notifications will not work.

jklare:
I totally agree with @punkle here, but i think the biggest issue is the fact that these are definitions. @svanzoest Are there any plans on refactoring these to resources? I would be happy to help with that.

svanzoest:
@jklare this is part of #414 and the [apache2@customresources](https://github.com/svanzoest-cookbooks/apache2/tree/customresources) branch has the current implementation. I am looking for feedback and some testing on that branch. I'd love to get your perspective on #414

jklare:
@svanzoest sounds great, i will look trough the changes in that branch and post my feedback in #414 (sry for responding so slowly, i was on holiday the last 2 weeks)

Issue #310, opened 2015-01-22
Can't restart apache, due to /var/lock/apache2 ACL issue


DennyZhang:
Hi there

Below is the error message of running chef update.
I've checked the code, /var/lock/apache2 is changed to root on purpose.
Do you know what's the problem? My OS is Ubuntu 12.04.

---- Begin output of /etc/init.d/apache2 start ----
STDOUT: * Starting web server apache2
/var/lock/apache2 already exists but is not a directory owned by www-data.
Please fix manually. Aborting.
   ...fail!
STDERR: 
---- End output of /etc/init.d/apache2 start ----
Ran /etc/init.d/apache2 start returned 1

drpebcak:
This doesn't happen when just running the apache2 cookbook. Can you post the full output of your chef run? Likely the nagios3 cookbook is doing something that apache2 does not expect.

DennyZhang:
@drpebcak 

https://github.com/viverae-cookbooks/apache2/blob/master/recipes/default.rb#L115-L125
   115	%W(
   116	  #{node['apache']['dir']}/ssl
   117	  #{node['apache']['cache_dir']}
   118	  #{node['apache']['lock_dir']}
   119	).each do |path|
   120	  directory path do
   121	    mode '0755'
   122	    owner 'root'
   123	    group node['apache']['root_group']
   124	  end
   125	end

If I comment off Line 117, I won't have the issue.
Above code change the ownership of /var/lock/apache2 to root, instead of www-data.

This looks like to be the root cause.

joshuacox:
I'm getting the same error, now don't laugh at the sprunge url that really did come back:

http://sprunge.us/FLaW

drpebcak:
Huh, I take it back. This must be something that has changed recently. I'm going to check if this also happens with apache2.4

drpebcak:
Looks like this only happens with apache2.2

svanzoest:
This got set because of #296.

Related Directives:
- [Apache 2.4: Mutex](http://httpd.apache.org/docs/2.4/mod/core.html#mutex)
- [Apache 2.2: LockFile](http://httpd.apache.org/docs/2.2/mod/mpm_common.html#lockfile)

joshuacox:
after reading through #296 I'm a bit confused as this was working just fine a few months ago?

Just to check I threw this in my default attributes:

```
    "apache": {
      "user": "www-data",
      "group": "www-data"
    },
```

[...]

PR #450, opened 2016-10-31
Ensure the lock_dir is owned by www-data for Apache 2.2 and 2.4 on Debian/Ubuntu

meineerde:
## Description

The default `apache2ctl` script for both Apache 2.2 and 2.4 creates the `lock_dir` in `/var/lock/apache2`. If the directory exists but is not owned by `www-data`, the script fails, similar to what is described in #310 and #312.

### Issues Resolved

This commit enhances the fix from #312 for Apache 2.4 in Debian/Ubuntu.

### Check List
- [ ] All tests pass. See https://github.com/svanzoest-cookbooks/apache2/blob/master/TESTING.md

### DIFF
--- a/recipes/default.rb
+++ b/recipes/default.rb
@@ -127,7 +127,7 @@ end
 
 directory node['apache']['lock_dir'] do
   mode '0755'
-  if node['platform_family'] == 'debian' && node['apache']['version'] == '2.2'
+  if node['platform_family'] == 'debian' && %(2.2 2.4).include?(node['apache']['version'])
     owner node['apache']['user']
   else
     owner 'root'


meineerde:
Note that the failing test is due to unrelated Foodcritic complaints.

As for testing, I'm not sure this warants new tests since the previous behavior is not tested either. This pull request just extends existing functionality from version 2.2 to version 2.4.

meineerde:
And now just that you merged it (thanks for that!) I find the typo...

`%(2.2 2.4)` is equal to `"2.2 2.4"`. So while it works fine with a string comparison, it is not entirely correct since it actually should compare to an Array, i.e `%w(2.2 2.4)`.

I just pushed another commit to my branch to fix this.

svanzoest:
You can not do a version check on apache.version like that as it is a flag. So I removed it entirely

meineerde:
Ah, and now, as I actually read the connected commit, it all makes sense. Thank you, keep up the good work!

PR #451, opened 2016-11-02
Add custom config to envvars file

chuhn:
## Description

We define custom environment variables in /etc/apache2/envvars ie. `ORACLE_HOME`. 
So either we need:
1. to extend this file with custom statements or 
2. else add an option to apache2::default to disable the management of the envvars file
3. or add a single statement to include an optional include not managed by apache2 cookbook (eg. `[ -r /etc/apache2/envvars.local ] && . /etc/apache2/envvars.local`?)

Below you find my minimal initial attempt. Is this the direction to go? Or would you prefer the 3rd variant?

### Check List
- [x] All tests pass. See https://github.com/svanzoest-cookbooks/apache2/blob/master/TESTING.md


chuhn:
OK. I decided to go for solution 3 and source a =/etc/apache2/envvars.local= if it exists.

Cheers,
	Christopher

svanzoest:
We should focus on consistency and therefore I think it should work just like `apache.sysconfig_additional_params`.

chuhn:
After a long pause I updated my PR to work like `sysconfig_additional_params`.

No unit tests seem to exist for the envvars file yet.

damacus:
Hey! 

Thanks,
Dan

damacus:
So this should be able to be fixed in a completely different way now.  So I'm going to close this down and suggest a rebase from master.

PR #453, opened 2016-11-11
Fixes to attributes

tas50:
Some code optimizations and bug fixes in the attributes

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -80,8 +80,8 @@ default['apache']['root_group'] = 'root'
 default['apache']['default_site_name'] = 'default'
 
 # Where the various parts of apache are
-case node['platform']
-when 'redhat', 'centos', 'scientific', 'fedora', 'amazon', 'oracle'
+case node['platform_family']
+when 'rhel', 'fedora'
   if node['platform'] == 'amazon'
     if node['apache']['version'] == '2.4'
       default['apache']['package'] = 'httpd24'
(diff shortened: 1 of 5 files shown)


definitions/apache_module.rb
```diff
@@ -17,7 +17,7 @@
 # limitations under the License.
 #
 
-define :apache_module, :enable => true, :conf => false, :restart => false do # rubocop:disable Metrics/BlockLength
```
svanzoest: This will break in newer versions of rubocop, the real way to fix this is to refactor as part of #414 .

recipes/mod_php.rb
```diff
@@ -25,7 +25,7 @@
   Chef::Log.warn("Currently the apache2 cookbook is configured to use the '#{node['apache']['mpm']}' MPM")
 end
 
-case node['platform_family'] # rubocop:disable Style/MultilineIfModifier
```
svanzoest: This will break in newer versions of rubocop, the real way to fix this is to refactor as part of #414 .

attributes/default.rb
```diff
@@ -80,8 +80,8 @@
 default['apache']['default_site_name'] = 'default'
 
 # Where the various parts of apache are
-case node['platform']
-when 'redhat', 'centos', 'scientific', 'fedora', 'amazon', 'oracle'
```
svanzoest: these are explicitly `platform` and not `platform_family` as we do not have `platform_family` in serverspec, so we can't match platform fixtures and this fails. It is much easier to keep this this way until we add platform_family support to the integration test fixtures.

attributes/default.rb
```diff
@@ -48,9 +48,9 @@
     when 'ubuntu'
       node['platform_version'].to_f >= 13.10 ? '2.4' : '2.2'
     when 'linuxmint'
-      node['platform_version'].to_i >= 16 ? '2.4' : '2.2'
+      node['platform_version'].to_i >= 16 ? '2.4' : '2.2' # this can go away 4/17
```
svanzoest: why? what is happening with linuxmint?

tas50: I should clarify there that 16 goes EOL 4/17

definitions/apache_module.rb
tas50: #454 addresses this with cookstyle in Travis

svanzoest: it would still fail rubocop

tas50: If you use rubocop then yes it would fail

PR #454, opened 2016-11-11
Test in Travis with Chef-DK and a Rakefile

tas50:
Using ChefDK avoids having to match Ruby releases and Gem releases to match Chef tooling.


tas50:
This is going to fail until the extra cops are removed. That fix is in #453

.rubocop.yml
```diff
@@ -24,3 +24,5 @@ TrivialAccessors:
   ExactNameMatch: true
 Style/RegexpLiteral:
   Enabled: false
+Style/ExtraSpacing:
+  Enabled: false
```
svanzoest: what is the reasoning for turning this off?

tas50: It attempts to change lines in the Gemfile without this off

Gemfile
```diff
@@ -8,7 +8,7 @@ gem 'berkshelf', '~> 5.1', '>= 5.1.0'
 
 group :unit do
   gem 'foodcritic',       '~> 8.1', '>= 8.1.0'
-  gem 'rubocop',          '~> 0.45', '>= 0.45.0'
+  gem 'cookstyle'
```
svanzoest: I don't understand the motivation for cookstyle, why not just add additional rules to rubocop?

tas50: Cookstyle lets you avoid all rules that make zero sense in a cookbook. It simply avoids users having to start with a 30 lines .rubocop.yml when they create a new cookbook. As time goes on we expect we'll add the logic of foodcritic into cookstyle as rubocop plugins so it just becomes the linting for both Chef DSL and Ruby.

zuazo:
Travis/RuboCop error (it seem easy to fix):

```
definitions/apache_module.rb:20:78: W: Unnecessary disabling of Metrics/BlockLength (did you mean Metrics/BlockNesting?).
define :apache_module, :enable => true, :conf => false, :restart => false do # rubocop:disable Metrics/BlockLength
                                                                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
recipes/mod_php.rb:28:30: W: Unnecessary disabling of Style/MultilineIfModifier (did you mean Style/MultilineIfThen?).
case node['platform_family'] # rubocop:disable Style/MultilineIfModifier
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

svanzoest:
That applies to robocop 0.45.

.travis.yml
```diff
@@ -1,19 +1,35 @@
...
+branches:
+  only:
+    - master
+
+# Ensure we make ChefDK's Ruby the default
 before_script:
-  - bundle exec berks
+  - eval "$(/opt/chefdk/bin/chef shell-init bash)"
+  # We have to install chef-sugar for ChefSpec
+  - /opt/chefdk/embedded/bin/chef gem install chef-sugar
```
svanzoest: why not leverage the gem file here?

tas50: We can actually nuke this part. This came over from the cookbook I copied it from

PR #455, opened 2016-11-11
openSUSE Leap has it's own platform name

tas50:
This is a unique platform in the suse platform family


zuazo:
Related ohai issue: https://github.com/chef/ohai/pull/784

PR #470, opened 2017-02-21
Remove support for EOL Fedora / FreeBSD

tas50:
FreeBSD 9 is EOL and Fedora 17 is way EOL. Both of these platforms are Apache 2.4 only now.


attributes/default.rb
```diff
@@ -75,8 +75,6 @@
       '2.4'
     end
   when 'freebsd'
-    node['platform_version'].to_f >= 10.0 ? '2.4' : '2.2'
-  else
```
shortdudey123: Removing the default case?

tas50: Line 80 is 2.4

shortdudey123: i mean you are removing the `else`, just making sure thats what you want since this will have no default anymore

Issue #361, opened 2015-06-16
Heads up: Amazon Linux separate from RedHat in serverspec os object now


yoshiwaan:
https://github.com/serverspec/specinfra/releases

When running serverspec os[:family] reports amazon linux to be 'amazon' now instead of 'redhat'

This means that [apache2/test/fixtures/serverspec_helper.rb](https://github.com/svanzoest-cookbooks/apache2/tree/master/test/fixtures/serverspec_helper.rb) will need updating with something like the following:

```
osmapping = {
  'amazon' => {
    :platform_family => 'rhel',
    :platform => 'amazon',
    :platform_version => '2014.09'
  },
```

I would say platform_version has to be 2014.09 as unless fauxhai has been updated pretty recently 2015.03 isn't supported yet

yoshiwaan:
Hmm, it's a little trickier as in supported versions for Chefspec 2014.09 is used but when running serverspec then the logic in serverspec_helper.rb (below) sets the version to the latest amazon linux if you're using it (2015.03).
```
  # Hack: avoid freebsd to just use a release of 10 and force 10.0 instead to match chefspec
  if os[:release] && os[:family] != 'freebsd'
    ohaistub[:platform_version] = os[:release]
  else
    ohaistub[:platform_version] = osmapping[os[:family]][:platform_version]
  end
```
So either two json attribute files need to exist in test/fixtures/platforms/amazon or serverspec_help.rb needs updating to force 2014.09, of which the second is probably the better option.

svanzoest:
Thanks for the heads up. We can also put in a request for 2015.03 be added to [fauxhai](https://github.com/customink/fauxhai)

svanzoest:
We now moved to inspec #460, so not directly related but also addressed with #478 .

PR #478, opened 2017-04-11
support amazon as a platform_family as per chef/ohai#971

svanzoest:
## Description

With Chef 13 the amazon kitchen test fails entirely, because it is looking for `platform_family` of `amazon` instead of `rhel` and therefore all the platform specific code doesn't run.

This is tested against chef 13 and based on a review chef 12 should continue to work.

### Issues Resolved

Related to #361 

### Check List
- [X] All tests pass. See https://github.com/chef-brigade/apache2/blob/master/TESTING.md


svanzoest:
Food critic FC024 seems to be failing as maybe https://github.com/acrmp/foodcritic/pull/550 is not released yet?

PR #488, opened 2017-06-13
Sysconfig issue486

klarakarl:
## Description

When cookbook is configured to install httpd24 package wrong config file is generated

The generated is 
   /etc/sysconfig/httpd24
Instead of 
  /etc/sysconfig/httpd


damacus:
@svanzoest are we OK to merge this one in?

svanzoest:
@damacus looks okay to me assuming all tests pass.

PR #491, opened 2017-07-06
Run integration tests in Travis + other testing updates

shoekstra:
- Update test framework to be in line with other SC cookbooks.
- Aim to test all mentioned supported OSes


.kitchen.yml
```diff
@@ -1,117 +1,120 @@
...
-- name: freebsd-10.3
-- name: amazon
-  driver:
-    box: mvbcoding/awslinux
-- name: opensuse-13.2
+  - name: amazonlinux
+    driver_config:
+      box: mvbcoding/awslinux
+  - name: centos-7.3
+  - name: debian-8.6
```
tas50: We should get Debian 9 in here as well

.kitchen.yml
```diff
@@ -1,117 +1,120 @@
...
+    run_list:
+    - recipe[apt::default]
+  - name: fedora-25
+  - name: freebsd-10.3
+  - name: opensuse-13.2
```
tas50: 13.2 is EOL and should get replaced with opensuse-leap-42.2

.travis.yml
```diff
@@ -1,28 +1,48 @@
-# Use Travis's cointainer based infrastructure
-sudo: false
+sudo: required
+dist: trusty
 
 addons:
   apt:
     sources:
-      - chef-current-precise
+      - chef-stable-trusty
```
tas50: We can test on current again now that we don't delete the promoted build out of current when we take a build to stable.

.travis.yml
```diff
@@ -1,28 +1,48 @@
...
+  - INSTANCE=event-ubuntu-1604
+
 before_script:
+  - sudo iptables -L DOCKER || ( echo "DOCKER iptables chain missing" ; sudo iptables -N DOCKER )
   - eval "$(/opt/chefdk/bin/chef shell-init bash)"
-  # We have to install chef-sugar for ChefSpec
-  - /opt/chefdk/embedded/bin/chef gem install chef-sugar
-
-script:
   - /opt/chefdk/embedded/bin/chef --version
```
tas50: /opt/chefdk/embedded/bin/ should get removed everywhere in this config now

spec/support/shared_examples_for_apache_module.rb
```diff
@@ -10,7 +10,6 @@
   module_name = a2module
   module_enable = true
   module_conf = a2conf
-  module_identifier = "#{module_name}_module"
```
tas50: ?

shoekstra: cookstyle was complaining about it being a useless assignment so I removed it.

tas50: works for me then :)

shoekstra:
LGTM!  Awesome work.

damacus:
Apart from that one minor comment that looks great!

Berksfile
```diff
@@ -3,11 +3,6 @@ source 'https://supermarket.chef.io'
 metadata
 
 group :integration do
-  cookbook 'apt'
   cookbook 'fqdn', git: 'https://github.com/drpebcak/fqdn-cookbook.git'
```
shoekstra: Looks like  this isn't being used and can be removed?

tas50: we include fqdn in the runlist in testing. It's probably something we could wrap into the setup recipe there though

PR #492, opened 2017-07-07
Remove cloud kitchen config for centos 5 / ubuntu 12.04

tas50:
This whole file should probably go away, but these need to go for sure

Signed-off-by: Tim Smith <tsmith@chef.io>


tas50:
Lets do #491 instead

PR #495, opened 2017-07-10
Simplify the cookbook by remove Apache 2.2 support

tas50:
There are two major platforms that have non-EOL releases utilizing Apache 2.2 and neither of these were listed as a tested/supported platform in our readme:

RHEL 6
Debian 7

Signed-off-by: Tim Smith <tsmith@chef.io>


shoekstra:
Yes 👍

PR #496, opened 2017-07-11
Fix opensuse support

shoekstra:
Not pretty but fixes opensuse default suite.

### DIFF
--- a/recipes/mod_cgi.rb
+++ b/recipes/mod_cgi.rb
@@ -19,6 +19,11 @@
 #
 
 if node['apache']['mpm'] == 'prefork'
+  link '/usr/lib64/apache2/mod_cgi.so' do
+    to '/usr/lib64/apache2-prefork/mod_cgi.so'
+    only_if { node['platform_family'] == 'suse' }
+  end
+
   apache_module 'cgi'
 else
   Chef::Log.warn "apache::mod_cgi. Your MPM #{node['apache']['mpm']} seems to be threaded. Selecting cgid instead of cgi."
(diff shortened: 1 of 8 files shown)


recipes/mod_cgi.rb
```diff
@@ -19,7 +19,9 @@
 #
 
 if node['apache']['mpm'] == 'prefork'
-  apache_module 'cgi'
+  apache_module 'cgi' do
+    module_path '/usr/lib64/apache2-prefork/mod_cgi.so' if node['platform_family'] == 'suse'
```
svanzoest: This seems to be applied for all platforms, seems like the if statement is not being applied.
```
 3) apache2::mod_cgi on Ubuntu 14.04 it should behave like an apache2 module creates <apache_dir>/mods-available/cgi.load
     Failure/Error:
       expect(chef_run).to create_file("#{chef_run.node['apache']['dir']}/mods-available/#{module_name}.load").with(
         content: "LoadModule #{module_name}_module #{chef_run.node['apache']['libexec_dir']}/#{module_filename}\n",
         mode: '0644'
       )
     
       expected "file[/etc/apache2/mods-available/cgi.load]" to have parameters:
     
         content "LoadModule cgi_module /usr/lib64/apache2/mod_cgi.so\n", was "LoadModule cgi_module /usr/lib64/apache2-prefork/mod_cgi.so\n"
       Diff:
       @@ -1,2 +1,2 @@
       -LoadModule cgi_module /usr/lib64/apache2/mod_cgi.so
       +LoadModule cgi_module /usr/lib64/apache2-prefork/mod_cgi.so
```

PR #504, opened 2017-07-31
Testing updates

tas50:
Use the dokken images
Use the standard test doc


damacus:
ohai data?

tas50:
ChefDK has been updated so the specs pass now

PR #505, opened 2017-08-15
Match mod_php .conf to module name

petracvv:
## Description

Since the apache module name is php5 and the module configuration is deployed as php.conf, enabling the php5 module does not also enable the appropriate configuration. This results in apache not properly displaying and executing .php files since various important configuration items are missing (including the .php extension handler).

This behavior was seen on CentOS 7. The incorrect file is also managed by Chef on Ubuntu but the libapache2-mod-php5 package there includes a php5.conf file so it all appeared to be working. 

This pull request matches the mod_php configuration filename to the module name specified in the cookbook attributes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 This file is used to list changes made in each version of the apache2 cookbook.
 
+## v5.0.1 (2017-09-01)
+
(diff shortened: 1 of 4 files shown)


petracvv:
Eh.. Messed up my rebase

PR #510, opened 2017-09-08
apache_conf resource: pass params as varaibales to the template

InformatiQ:
## Description

in the apache_conf resourc, pass params as varaibales to the template


damacus:
Hey! 

Thanks,
Dan

damacus:
Hi there! 

you might now be looking to do this to the apache2_conf resource that's currently on master. 

Thanks! 
Dan

PR #511, opened 2017-09-09
Fix guard on service provider

alonant:
## Description

Fix guard on service provider
Bash guard causes chef "Why run" to fail since #{node['apache']['binary']} does not exist yet.


svanzoest:
Why does `apache.binary` not exist? It should exist after the package
install completes.

alonant:
Hi @svanzoest 
This is an issue only during WHY RUN mode of the Chef client.
Since no binaries are installed in that mode the guard causes the service provider to fail during compile time as Chef evaluates the guard and runs the bash code. The bash code fails since the file is not found.

My PR comes to fix that problem.

Regarding the tests I don't see why my change has "unstubbed" the guard.
The tests are failing on the lint and unit tests and they appear to be failing on all platforms even when the previous installation tests pass.

damacus:
@svanzoest I'm going to pull this one in as it doesn't break anything and adds why-run compatibility.

damacus:
@alonant if you could update the tests, we'll pull this in
Thanks!

damacus:
Closing due to inactivity. 

Please reopen this once the changes have been made!

PR #518, opened 2017-12-20
more reliable MPM installation on SUSE

375gnu:
## Description

In (Open)SUSE it's possible to install different MPM packages but switching between them is very unreliable. By default prefork variant is installed but if install later worker prefork will be used still and if you remove prefork package you'll have a dangling symlink `/usr/sbin/httpd -> /usr/sbin/httpd-prefork`. This is exactly what you have with this cookbook if you don't want prefork MPM.

### Issues Resolved

### Check List
- [~] All tests pass. See https://github.com/sous-chefs/apache2/blob/master/TESTING.md
- [+] New functionality includes testing.
- [+] New functionality has been documented in the README if applicable

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ This cookbook uses many attributes, broken up into a few different kinds.
 
 In order to support the broadest number of platforms, several attributes are determined based on the node's platform. See the attributes/default.rb file for default values in the case statement at the top of the file.
 
-- `node['apache']['package']` - Package name for Apache2
+- `node['apache']['package']` - Package name for Apache2, ignored on (open)suse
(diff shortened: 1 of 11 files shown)


damacus:
Hey! 

Thanks,
Dan

damacus:
Hi there! 

we you might have noticed we've done a fair bit of work on apache2 in the last month. you might want to do a bit of digging on the current master branch to see how MPM support now works. 

It definitely doesn't fix this case. But I've love to see a bunch of tests and hopefully a passing kitchen suite for suse if you're up for that?

Closing this PR now as I'm suggesting a rebase from master 👍 

Thanks!

Issue #517, opened 2017-11-22
How to install the specfic version of the apache using this cookbook as ther is no attribute defined in the cookbook to provide the version details


samakode:
### Cookbook version
5.0.1

### Chef-client version
12.19.36

### Platform Details
Cent os 7.1 rhel family.  Azure cloud

### Scenario:
Trying to install the fixed version of apache 2.4.17

### Steps to Reproduce:
How to specify a specific version of apache

### Expected Result:
Fix the version of apache to 2.4.17 even if the new version is available via node attribute. 

### Actual Result:
After running this cookbook apache is upgraded to latest version automatically and overwrite the f8xed version that i am trying to install.

damacus:
Hi there!

we've recently done a sizable refactor, but would be happy to receive a PR to make the tunable. 

Thanks,
Dan

josephholsten:
generally we like to solve this at the package-management level: configure a yum repo and pin the package number. We would definitely love some documentation explaining this pattern, but it's somewhat outside the scope of this cookbook.

PR #519, opened 2017-12-20
Return chef 11 support

375gnu:
## Description

Some enterprises still use chef 11, this PR partially returns its support, bits not covered with it are the matter of #518 (packages removal on suse).

### Check List

### DIFF
--- a/metadata.rb
+++ b/metadata.rb
@@ -1,9 +1,8 @@
 name 'apache2'
-source_url 'https://github.com/sous-chefs/apache2'
-issues_url 'https://github.com/sous-chefs/apache2/issues'
+source_url 'https://github.com/sous-chefs/apache2' if respond_to?(:source_url)
+issues_url 'https://github.com/sous-chefs/apache2/issues' if respond_to?(:issues_url)
 maintainer 'Sous Chefs'
 maintainer_email 'help@sous-chefs.org'
-chef_version '>= 12.1' if respond_to?(:chef_version)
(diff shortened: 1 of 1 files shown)


thommay:
i'm 👎 on this - Chef 11 is not supported, and in April Chef will be removing it from download sites, so there'll be no way of testing it.

damacus:
Hi @375gnu thanks for  the PR. We won't be pulling in any support for Chef 11 builds. We look to keep current, which means around 18months of support for each Chef client version (possibly more if you're [un]lucky!) 

If you wish to use a version of the apache2 cookbook that supports Chef 11, you'll have to go back a fair bit. However in this case that is what version pinning is for.

375gnu:
On 12/20/17, Dan Webb <notifications@github.com> wrote:

> Hi @375gnu thanks for  the PR. We won't be pulling in any support for Chef
> 11 builds. We look to keep current, which means around 18months of support
> for each Chef client version (possibly more if you're [un]lucky!)
>
> If you wish to use a version of the apache2 cookbook that supports Chef 11,
> you'll have to go back a fair bit. However in this case that is what version
> pinning is for.

Unfortunately previous versions doesn't support SUSE12.

Version 5 doesn't have any Chef 12 features that prevent Chef 11 usage
(actually it has one resource "package" used with several arguments
but that recipe is buggy and addressed in PR #517).

So it's a pity to hear that you don't want to support real-world installations.

PR #530, opened 2018-03-26
Detect 64-bit architectures (fixes #529)

### LINKED ISSUE #529: 32-bit ARM is flagged as a 64-bit architecture
### Cookbook version
5.0.1

### Chef-client version
13.8.5

### Platform Details
CentOS 7 on 32-bit ARM

### Scenario:
Running the `default` recipe.

### Steps to Reproduce:
The minimal Chef run list `recipe[apache2]` when bootstrapped against a 32-bit ARM machine is sufficient to reproduce.

### Expected Result:
Apache httpd should install and start up.

### Actual Result:
Apache httpd installs but cannot start, because `/etc/httpd/modules` is erroneously linked to `/usr/lib64/httpd/modules`. It should be linked to `/usr/lib/httpd/modules`, as this is a 32-bit architecture.

don-code:
## Description
The current behavior on RHEL and compatibles is to assume a machine is 64-bit unless a regex for i386/i686 is matched. This causes alternative architectures (e.g. 32-bit ARM) to be misdetected as 64-bit.

This change attempts to better detect 64-bit architectures:
* Any machine type including "64" is assumed to be 64-bit. This is generally the case for supported architectures, e.g. "x86_64", "aarch64", "ppc64", etc.
* Some known 64-bit machine types without "64" are explicitly matched, e.g. "armv8l", "s390x".
* Anything else is assumed to be 32-bit (e.g. i686, armv7l)

### Issues Resolved

#529 (32-bit ARM is flagged as a 64-bit architecture)

### Check List
I ran an abbreviated set of tests against CentOS (as this covers the RHEL/Fedora/Amazon platform code path). I did not test the SuSE path as there is no Test Kitchen profile for it. I additionally tested against CentOS on an armv7l box.
I have not added tests for the new non-x86_64 cases.
No new functionality is included.

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -18,6 +18,19 @@
 # limitations under the License.
 #
 
+# Gets the libdir for a given CPU architecture
+def lib_dir_for_machine(arch)
+  if arch =~ /64/ || %w(armv8l s390x).include?(arch)
+    # 64-bit architectures
+    # (x86_64 / amd64 / aarch64 / armv8l / etc.)
+    '/usr/lib64'
+  else
+    # 32-bit architectures
+    # (i686 / armv7l / s390 / etc.)
+    '/usr/lib'
+  end
+end
+
 default['apache']['mpm'] =
   case node['platform']
   when 'ubuntu', 'linuxmint'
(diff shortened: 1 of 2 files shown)


damacus:
The test failures are something we need to fix with they was we use inspec. 

Thanks for the PR!

PR #533, opened 2018-04-02
Adding version to support non-apt based systems

eanrollings:
## Description

The `default_release` attribute is not supported on yum-based package managers (only apt).  Adding use of `version` to specify exact version of Apache to install.


damacus:
Hey! 

Thanks,
Dan

damacus:
Hi there, I don't think this is needed on master anymore. If you want to take a quick look at master, and see how  that works for you. 

What platform are you working on, and what bug does  this fix?

PR #534, opened 2018-06-19
Update apache2.conf.erb

kbuchsicf:
Uncomment conditionals around User and Group statements - as they are undefined without the module unixd loaded. In addition, define DocumentRoot to match the attribute node['apache']['docroot_dir']. This means the server will come up in a working configuration.

## Description

### Issues Resolved

### Check List

### DIFF
--- a/templates/default/apache2.conf.erb
+++ b/templates/default/apache2.conf.erb
@@ -40,10 +40,10 @@ MaxKeepAliveRequests <%= node['apache']['keepaliverequests'] %>
 #
 KeepAliveTimeout <%= node['apache']['keepalivetimeout'] %>
 
-#<IfModule unixd_module>
-User <%= node['apache']['user'] %>
-Group <%= node['apache']['group'] %>
-#</IfModule>
+<IfModule unixd_module>
+  User <%= node['apache']['user'] %>
+  Group <%= node['apache']['group'] %>
+</IfModule>
 
 # Sets the default security model of the Apache2 HTTPD server. It does
 # not allow access to the root filesystem outside of /usr/share and <%= node['apache']['docroot_dir'] %>.
@@ -66,6 +66,8 @@ Group <%= node['apache']['group'] %>
         AllowOverride None
         Require all granted
 </Directory>
+
+DocumentRoot <%= node['apache']['docroot_dir'] %>
 
 <Directory <%= node['apache']['docroot_dir'] %>>
         Options Indexes FollowSymLinks


templates/default/apache2.conf.erb
```diff
@@ -40,10 +40,10 @@ MaxKeepAliveRequests <%= node['apache']['keepaliverequests'] %>
...
 KeepAliveTimeout <%= node['apache']['keepalivetimeout'] %>
 
-#<IfModule unixd_module>
-User <%= node['apache']['user'] %>
-Group <%= node['apache']['group'] %>
-#</IfModule>
+<IfModule unixd_module>
+  User <%= node['apache']['user'] %>
+  Group <%= node['apache']['group'] %>
+</IfModule>
```
svanzoest: 👍 This looks good for apache 2.4+ as [User and Group](https://httpd.apache.org/docs/2.4/mod/mod_unixd.html#user) are now part of `unixd`. The reason it was commented out was because on apache 2.2, the directives `User` and `Group` where part of core and there was no `unixd` module so this would configuration change would have failed.

templates/default/apache2.conf.erb
```diff
@@ -67,6 +67,8 @@ Group <%= node['apache']['group'] %>
         Require all granted
 </Directory>
 
+DocumentRoot <%= node['apache']['docroot_dir'] %>
+
```
svanzoest: I am 👎 on this.

We explicitly did not set the global `DocumentRoot` directive for security reasons, as it is best practice to use the `web_app` definition to create a `<VirtualHost>` which would include the `DocumentRoot` directive.

By including it globally, we would not have a way to turn it off and there is very little reason these days to enable it on a production apache httpd installation.

kbuchsicf: OK, if this is the intention, then why not remove the <Directory> block on the same? In addition, I'm not sure of the purpose of the <Directory> on /usr/share/doc - seems like a loose end and confusing.

kbuchsicf:
I made the changes discussed in my comments. I threw in a few other changes. One makes it easier to rename the cookbook, removing the cookbook name from self references.

templates/default/apache2.conf.erb
kbuchsicf: I suppose this is a debatable point. If the main direction you want uses to go in is to create virtual sites, then maybe it is left as documentation (comments). For rhel/centos distros the default DocumentRoot is /etc/httpd/htdocs. So, this <Directory /var/www/html> can be very confusing. I went through that when learning the cookbook.

Maybe even better than what I did (replace with comments in httpd.conf) would be to also create an example Virtual Site configuration file, which is all comments, but serves as a handy reference to get people up and running. Put that file in place in sites-available and you would give the users something really helpful.

damacus:
Hey @kbuchsicf could you change just the conf file and I'll merge this into master. We're working on loads of other improvements to this cookbook right now and we're likely to make more merge conflicts (as you can see!)

Thanks very much for this contribution, it's really appreciated!

kbuchsicf:
@damacus - I just want to be sure I'm doing what you want. Are you looking for me to make some changes to the conf file as part of this PR or are you looking for a different PR which only makes changes to the conf file. @svanzoest indicated he disagreed with my changes to the apache2.conf.erb template. I did alter the file so it doesn't define a DocumentRoot as my original commit did - but added some additional comments to explain what I thought was needed. I thought this would have resolve svanzoest's concern.

damacus:
Different pr please. I think it'll be easier. If you just change the conf file and the related templates I'll merge that in ASAP. It probably won't be released until I'm happy with the refactoring thats going on right now. Hence all the conflicts

PR #543, opened 2018-07-30
service name helper

damacus:
## Description

Add service_name helper

### Check List


libraries/helpers.rb
```diff
@@ -3,18 +3,25 @@ module Cookbook
     module Helpers
       def apache_binary
         case node['platform_family']
-        when 'rhel', 'fedora', 'amazon', 'arch'
-          '/usr/sbin/httpd'
-        when 'suse'
+        when 'rhel', 'fedora', 'amazon', 'arch', 'suse'
```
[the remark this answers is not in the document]
miedward: There still has to be a case so I don't know that it maters a lot what the default is.  I guess it depends on what is the most "popular" server platform for apache is.

recipes/default.rb
```diff
@@ -200,10 +200,8 @@
   end
 end
 
-apache_service_name = node['apache']['service_name']
-
 service 'apache2' do
-  service_name apache_service_name
+  service_name service_name
```
miedward: Can you call a service_name with a variable named service_name?  That is super meta if so.  Maybe make it "#{service_name}" to make it clearer what that one is a string variable and one is a key word?

damacus: The helper would need to be called something else really., the interpolation would fail the rubocop

PR #547, opened 2018-08-20
Allow httpd -t timeout to be configurable

skadz:
## Description
This allow the timeout for httpd -t to be configurable.  On slower machines, this regularly times out for me, so it would be nice if I could make it a bit longer.

### Issues Resolved
None

### Check List
- [N/A] New functionality includes testing.

Not sure if this should really go in to the README?  It maybe should, but seems like an "Advanced Setting" and I don't see a section for that.  Can I get some guidance here?  Thanks!


damacus:
Nice one.

I'll look at shipping this, this week. If I haven't done it by Thursday pester me!

PR #548, opened 2018-08-21
Add apache_conf_dir helper

skadz:
Update httpd.conf location

## Description
httpd.conf needs to be in the conf subdir on rhel family (and I am guessing others).  This adds a new helper that points to the conf dir, so that the httpd.conf file is created in the correct location.

### Issues Resolved
Didn't start an issue, but this caused my use of this to fail (ports.conf wasn't being read because the httpd.conf was being written in the wrong dir)

### Check List

### Notes
Guessing you probably want to clean up the attributes file too, since many of them are not used any more (like default['apache']['conf_dir'])


damacus:
This is great, could you throw in a test for this please? 

Example here 👍 feel free to break that out into it's own file. 

https://github.com/sous-chefs/apache2/blob/master/spec/libraries/helpers_spec.rb#L156

skadz:
Tests added, sorry for missing that the first time!

damacus:
This is excellent thank you.

FYI, the current helpers were much easier to spec out in unit tests first then make the changes.

PR #550, opened 2018-08-21
Fix ports.conf location and how its set up

skadz:
## Description
This was a fun one to figure out :)

ports.conf was created in conf-available, but not enabled by the default recipe
ports.conf was create in the apache_dir ONLY by mod_ssl

Because of this, the include in the default httpd.conf only worked if you enabled mod_ssl

So, what I did was to enable ports in the default recipe and then update it in mod_ssl if the defaults are kept.

I also updated the tests to match this.

May not be the right way, but it gets it working with just default.  Totally open to suggestions if there is a better way.

### Issues Resolved
httpd -t fails with default recipe

### Check List
- [X] New functionality has been documented in the README if applicable


damacus:
Hah! Brilliant!

Do you fancy helping out with the refactor? You seem like you're pretty handy with the apache cookbook!

PR #552, opened 2018-08-27
Restore ports.conf management to the default location

scassiba:
## Description

The ports.conf that gets emitted by the template resource should live at #{apache_dir}/ports.conf as the package also drops off its own ports.conf in the same location. Continue to manage the file at the default path.

This is a partial revert of #550 to move the resource management to the default recipe.

### Issues Resolved

### Check List
- [x ] All tests pass. See https://github.com/sous-chefs/apache2/blob/master/TESTING.md


damacus:
Thanks dude. I'll get this released tomorrow

ideaship:
@damacus @scassiba You guys may want to tag this change, it is easier to pick up that way.

PR #610, opened 2019-03-05
Avoid name collisions: Disable including helpers module in Recipe/Resource class

bottkv488:
# Description
i.e. docker fails because of `libexec_dir` name collision in `librarie/helpers.rb`.
* fixed include
* fixed nested array
* added test stub

## Issues Resolved

## Check List

- [x ] All tests pass. See https://github.com/sous-chefs/apache2/blob/master/TESTING.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,9 @@ This file is used to list changes made in each version of the apache2 cookbook.
 
 ## Unreleased
 
+- Use `declare_resource` in `apache2_module` 
 - Add default value to apache_2_mod_proxy
+
 
 ## v6.0.0 (tbc)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,8 @@ This file is used to list changes made in each version of the apache2 cookbook.
 
 ## Unreleased
 
-- Use `declare_resource` in `apache2_module` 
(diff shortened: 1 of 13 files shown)


damacus:
Hey thanks for this, I don't think this is the correct approach to fixing this. 

I'll get some time shortly to release without these at all.

bottkv488:
@damacus sorry to hear that, because it's a real blocker for us:
After all it is a released cookbook, and it breaks docker as soon as it's in the runlist ...

damacus:
I'll have this fixed in the next week or so. It'll be a major release. But I'm pretty happy it should be fixed.

damacus:
Fixed by #611

bottkv488:
NP, just wanted to help ;)

I guess you still might want to fix
https://github.com/sous-chefs/apache2/blob/6f3502150f3cc6c5facb427d9c024280e1909239/libraries/helpers.rb#L269
?
(That's why I added the new spec test, because the current one didn't catch the nested array)

damacus:
Really appreciate the help here. It's honestly amazing.

I thought this should have been covered by the current default_modules_spec.rb no? 🤔 (generally don't know the answer to this one!)

bottkv488:
I did not dig any further, my first guess would be that the log resource flattens the array?

PR #617, opened 2019-03-28
Fixed Mutex directive commented out

### LINKED ISSUE #616: Mutex directive commented out?
I've got a problem with my apache2 server. It randomly crashes. See for a similar issue this question: https://stackoverflow.com/questions/50652808/apache-shutdown-couldnt-grab-mutex. I think the problem started when I changed the user and group of apache2, but I'm not sure.

In `apache2/templates/apache2.conf.erb` I see the `Mutex` directive is commented out. Is this by design? If so, a comment explaining why it is commented out would be usefull.

stephanvierkant:
# Description

The `Mutex` was commented out, which makes no sense since there is a default value.

## Issues Resolved

Fixes #616

## Check List

### DIFF
--- a/templates/apache2.conf.erb
+++ b/templates/apache2.conf.erb
@@ -7,7 +7,7 @@ ServerRoot "<%= @apache_dir %>"
 #
 # The accept serialization lock file MUST BE STORED ON A LOCAL DISK.
 #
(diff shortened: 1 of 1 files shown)


damacus:
Ah! This was/is an oversight on my part. Thanks for catching it!

PR #624, opened 2019-04-15
Increase minimum chef version

bplessis:
# Description

Resource property 'documentation' field has been added at 13.9 release, version should be bumped to that number at minima

## Check List


damacus:
I think we can just drag our heels for a couple of weeks on this and just bump it straight to 14.0

bplessis:
Shouldn't the question be "is there a useful and required feature of 14.x that we will use" ? :)

MarkGibbons:
This version doesn't work on 13, so I'm in favor of bumping the minimum version to 14.0.  The description settings on the resource properties didn't compile on 13.8.5.

damacus:
Not really :) 
We support "current Chef -1"

PR #626, opened 2019-04-16
resource/install.rb: Auto-convert the listen property to an Array when it's a type String

bplessis:
# Description

Install resource specify the listen property as a String or an Array, however <% @listen.each do |value| -%> will fail when listen is of type String, so do a inconditionnal upgrade to an Array before handling the value to the template.

## Check List

### DIFF
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -34,3 +34,6 @@ suites:
   - name: ssl
     run_list:
       - recipe[test::mod_ssl]
+  - name: ports
+    run_list:
+      - recipe[test::ports]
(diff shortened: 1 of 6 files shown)


kitchen-porter:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Please include a <a href="https://github.com/sous-chefs/line-cookbook/blob/master/CHANGELOG.md">CHANGELOG</a> entry.</td>
    </tr>
  </tbody>
</table>

</p>

damacus:
I think in this case we should just allow Arrays as it's a plural "ports" so the use should really expect an Array in this case.

bplessis:
Hi, didn't saw your previous message, i can redo the PR to remove the String option in this case

damacus:
Nah we should be good for now, I think(!)

We can simplify on just an Array at a later date.

PR #635, opened 2019-05-15
Add package name and version parameters to install resource

### LINKED ISSUE #614: Override package installed
# Brief Description

Is it no longer possible to override the package name in the Apache2 cookbook?

## Cookbook version

7.0.0

## Chef-client version

12 & 14

## Platform Details

CentOS7

## Scenario

I wish to install a later version of Apache 2.4 via the httpd24u package from IUS repo. But there doesn't seem to be a way to install a specific package instead of the cookbook defined package. Not using the apache2_install resource would achieve this aim, but then presumably I don't get the advantage of the cookbook managing the apache config?

## Steps to Reproduce

apache2_install 'default'

## Expected Result

httpd24u to be installed instead of httpd

## Actual Result

httpd installed

Okonos:
# Description

Allow override of package name and version in install resource.

If the change is approved and anybody finds it desirable I may supply it with unit tests (another context in [install_spec](../blob/master/spec/resources/install_spec.rb) I believe).

## Issues Resolved

#614 

## Check List


kitchen-porter:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Please include a CHANGELOG entry.</td>
    </tr>
  </tbody>
</table>

</p>

resources/install.rb
```diff
@@ -102,7 +112,9 @@
          description: 'Hash of additional sysconfig parameters to apply to the system'
 
 action :install do
-  package [apache_pkg, perl_pkg]
+  package [new_resource.apache_pkg, perl_pkg] do
```
danbuchal: Wouldn't this also attempt to install the perl_pkg with the apache version?

resources/install.rb
```diff
@@ -102,7 +112,9 @@
...
+    version [new_resource.apache_version, nil] unless new_resource.apache_version.empty?
```
danbuchal: I discovered an issue with this.  I believe the apache_version must also be applied for any module that must also install a package.

Example: `apache2_module 'ssl'`
This module installs the `mod_ssl` package and will upgrade Apache to the latest available.

Okonos: Good catch, thanks. Apologies for not testing it well enough.
I get the impression that it would be best solved with the use of attributes, but you don't use them here for some reason.
Alternatively installing everything in one place could be viable somehow,  the `package` resource is not used in many places, so perhaps refactoring the resources using it is a solution. Passing a property to them is also a way, I'm not a fan of that though, as it puts too much of a burden on an user and is easy to forget about and thus error-prone.
These are just loose thoughts. Let me know your feedback. Do you see any other possibilities?

danbuchal: I too am still learning the proper ways of handling the transition from recipe-based to resource-based cookbooks and I'm not sure of the best methods yet myself.  @damacus do you have any suggestions?

damacus:
I'm going to close this PR off for now as I think we'll be better off discussing the design in an open issue or slack, the opening a PR 

This way we save you find down the wrong path for weeks.

resources/install.rb
damacus: Switching from recipe based to resource based in one way means we provide a consistent interface to the user of the cookbook. All tunables should be pushed up to the resource interface instead of allowing the user to reach into the cookbook and change things that we can't test. 

If we're adding in a resource property that can change multiple internal values, or change how something is (for example) installed. Then we should have a test for that.

damacus: This looks like the correct way of pushing up defaults that we can override if needs be. And thankfully it also looks like you'e done an awesome job with renaming the library helper to `default_`, which helps maintainers in the future, who are looking for default values!

ramereth:
I'd feel much better if this included a test cookbook recipe which actually tested doing an install from the IUS repo as was the original reason for this as mentioned in #614.

ramereth:
> I'd feel much better if this included a test cookbook recipe which actually tested doing an install from the IUS repo as was the original reason for this as mentioned in #614.

@Okonos can you please address this?

Okonos:
> > I'd feel much better if this included a test cookbook recipe which actually tested doing an install from the IUS repo as was the original reason for this as mentioned in #614.
> 
> @Okonos can you please address this?

Sorry, I no longer work with Chef (it's been a few months now). Feel free to take my work over, but to be honest there's not much of it and, as the above discussion shows, it's far from complete.

ramereth:
@Okonos ok, I'll re-open this with a local branch and see if we can make those changes.

resources/install.rb
ramereth: I'll take care of this issue in another PR once this is merged

PR #646, opened 2019-08-01
Property mod_conf must be one of: Hash!

lambovg:
Property mod_conf must be one of: Hash!  You passed nil.

# Description

```ruby
pache2_install 'default' do
  mpm 'prefork'
  mpm_conf(
      startservers: 2,
      minspareservers: 2,
      maxspareservers: 5,
      maxrequestworkers: 60,
      maxconnectionsperchild: 1000
  )
  modules %w(status)
  mod_conf(
      status: {
          status_allow_list: %w(127.0.0.1 ::1 1.2.3.0/24),
          extended_status: 'On',
          proxy_status: 'Off'
      }
  )
end
```
without modules status I get error Property mod_conf must be one of: Hash!  You passed nil.

## Issues Resolved

Wrong documentation example

### DIFF
--- a/documentation/resource_apache2_install.md
+++ b/documentation/resource_apache2_install.md
@@ -51,6 +51,7 @@ apache2_install 'default' do
            maxrequestworkers: 4096,
            maxconnectionsperchild: 1024
           )
+  modules %w(status)
   mod_conf(
     status: {
       status_allow_list: %w(127.0.0.1 ::1 1.2.3.0/24),
--- a/documentation/resource_apache2_module.md
+++ b/documentation/resource_apache2_module.md
@@ -23,7 +23,7 @@ Enable the php5 module, which has a different filename than the module default:
 
 ```ruby
 apache2_module "php5" do
-  filename "libphp5.so"
+  mod_name "libphp5.so"
 end
 ```
 


damacus:
Brilliant! Thank you for this.

I wonder why you're not getting our updated circleci config that accepts the chef license 🤔

lambovg:
Hey, thank you :) What I should do to get it :)

damacus:
I think it just need rebasing on master. Give that a try, and if that doesn't work then we'll try something else

Okonos:
What version of the cookbook and what OS did you get this error on?

The example from documentation already has modules passed and in the pull request it is added for a second time. Also, reading the source of the current version (`7.1.1`), I can't see a way it could be reproduced, as even passing an empty list of modules will not break this. Am I missing something?

I tested multiple variants of the example: unchanged, without `modules` line and with empty modules passed in (`modules []`) and all worked fine on version `7.1.1`.

lambovg:
7.1.0 on amazon linux 2.

lambovg:
currently Im with 7.1.1 and I can try to remove modules to see what will happened.

lambovg:
Yep you are right, thanks for your efforts - with the version 7.1.1 is working 100%. 

However on amazon linux 2 I've encountered another issue - httpd reload doesn't work it timeouts and httpd needs to be restarted. Actually this happened right now when I run the recipe.
I know that there is a bug report on your side about that but it was closed.
Actually this was a blocker for me because on our prod environments we need to be able to reload apache and without apache2 recipe + our stack its reloads successfully.

However next two week Im on vacation and I will take a look at this issue because its force me not to use apache2 cookbook.

chef-client hangs
```
 change mode from '' to '0644'
      * execute[a2enmod systemd] action run
         - execute /usr/sbin/a2enmod systemd

     * service[apache2] action enable
       - enable service service[apache2]
     * service[apache2] action start
       - start service service[apache2]
     * service[apache2] action restart
       - restart service service[apache2]
```

/var/log/messages
```
Aug  9 14:44:19 ip-0-0-0-0 systemd: httpd.service reload operation timed out. Stopping.
```

httpd status
```
active: reloading (reload) since Fri 2019-08-09 14:38:18 CEST; 8min ago
     Docs: man:httpd.service(8)
```

basically its not possible to finish that blocks chef-client to finish.

lambovg:
Yep I've tested with clean install on amazon linux 2

```
ervice 'apache2' do
  extend Apache2::Cookbook::Helpers
  service_name lazy {apache_platform_service_name}
  supports restart: true, status: true, reload: true
  action :nothing
end

apache2_install 'default_install' do
  mpm 'prefork'
  mpm_conf(
      startservers: 2,
      minspareservers: 2,
      maxspareservers: 5,
      maxrequestworkers: 60,
      maxconnectionsperchild: 1000
  )
  mod_conf(
      status: {
          status_allow_list: %w(127.0.0.1 ::1),
          extended_status: 'Off',
          proxy_status: 'Off'
      }
  )
  log_dir '/var/log/apache2-default'
end

apache2_config 'apache2.conf' do
  max_keep_alive_requests 100
  keep_alive_timeout 75
  timeout 15
  server_name 'localhost'
end
```

and I'm able to reproduce apache reload issue.

damacus:
Closed in favour of #650

damacus:
Thanks for the PR! 

Looks like we missed this one off the migration to the new CircleCI config which fixes the Chef licence problem.

PR #664, opened 2019-11-14
Changes to ports.conf / apache2.conf should trigger a service reload

bplessis:
Changes to ports.conf / apache2.conf should trigger a service reload, currently nothing happens.


damacus:
hey looks like your markdown isn't valid, can you throw a carrage return before your list, then the integration tests can run 👍 

[Here's a link to the tool we use ](https://github.com/markdownlint/markdownlint)

damacus:
This has been release to the supermarket under 8.0.1

thanks!

PR #669, opened 2019-12-13
Simplify platform logic and use ::File.exist in conditionals

tas50:
Fix all issues from the latest cookstyle master

Signed-off-by: Tim Smith <tsmith@chef.io>

PR #670, opened 2020-01-06
Migrate to Github Actions

Xorima:
# Description

This commit will migrate the build system to github actions

Signed-off-by: Jason Field <jason@avon-lea.co.uk>

## Check List


.github/workflows/ci.yml
```diff
@@ -0,0 +1,67 @@
...
+        uses: actionshub/markdownlint@master
+
+  dokken:
+    needs: [mdl, yamllint, delivery]
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        os:
+          - 'debian-8'
+          - 'debian-9'
```
tas50: Can we swap debian 8 for 10 while we're here

.github/workflows/ci.yml
```diff
@@ -0,0 +1,67 @@
...
+          - 'centos-7'
```
tas50: and add centos 8

Xorima: wow that response time :D 

Let's see if it all works first then will make tweaks as required, got ideas for a couple other workflows as well around touched the kitchen files and not this file and touched a workflow so reviewer please ensure all changes are correct, your thoughts?

.github/workflows/branchcleanup.yml
```diff
@@ -0,0 +1,17 @@
...
+
+env:
+  NO_BRANCH_DELETED_EXIT_CODE: 0
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: jessfraz/branch-cleanup-action@master
```
damacus: There is litteraly a tick box in the github UI fo this now.

libraries/helpers.rb
```diff
@@ -82,8 +82,7 @@ def lib_dir
       end
 
       def libexec_dir
-        case node['platform_family']
-        when 'freebsd', 'suse'
+        if platform_family?('freebsd') || platform_family?('suse')
```
Xorima: please fix my tests that all go bang ... :(

.github/workflows/branchcleanup.yml
Xorima: problem is you need admin access and they haven't exposed it to the API, which means terraform cannot do it for us.

This at least makes it more obvious to people as to how we are doing it and does not require an admin remembering something (We already have that with sponsers, I don't want to be the bottleneck to lots of things)

damacus:
Looks like the default-dokken-debian-8 is actually running way more than debian-8 :/

Xorima:
good catch @damacus , looks like something isn't feeding in on the action properly: `/usr/bin/kitchen test -`, code is in actionshub org so if you want to take a look feel free else I will later tonight

Xorima:
this also includes fixes for #669

.github/workflows/ci.yml
Xorima: Will create as a separate project for others to do, lets keep this focused on the build system else scope creep gets painful

PR #696, opened 2020-06-30
Add mod_php resource

### LINKED ISSUE #576: Resource: apache2_mod_php
# Cookbook version

6.0.0

## Scenario:

As a user
I want to be able to configure mod_php via properties rather than attributes
So that I can interact with mods in the same way.

## Expected Result:

- [ ] apache2_mod_php is configurable via resource properties
- [ ] mod_php has unit tests 
- [ ] no attributes under node['apache']['mod_php'] exist

detjensrobert:
# Description

This PR adds a new `apache2_mod_php` resource to configure and enable PHP.

## Issues Resolved

Fixes #576.

## Check List

### DIFF
--- a/documentation/resource_apache2_mod_php.md
+++ b/documentation/resource_apache2_mod_php.md
@@ -10,7 +10,7 @@ This resource will call `_module` with the correct identifiers for you.
 
 ## Properties
 
-| Name         | Type   | Default                                     | Description                         |
-| ------------ | ------ | ------------------------------------------- | ----------------------------------- |
-| module_name  | String | `php#{node['php']['version'].to_i}_module`  | The name of the php module.         |
-| so_filename  | String | `apache_mod_php_filename`                   | The filename of the module object.  |
+| Name         | Type   | Default                      | Description                         |
+| ------------ | ------ | ---------------------------- | ----------------------------------- |
+| module_name  | String | `apache_mod_php_modulename`  | The name of the php module.         |
+| so_filename  | String | `apache_mod_php_filename`    | The filename of the module object.  |
--- a/documentation/resource_apache2_mod_php.md
+++ b/documentation/resource_apache2_mod_php.md
@@ -5,7 +5,6 @@ Enables apache2 module `mod_php`.
 This resource assumes that php has already been installed with the [`php` cookbook](https://github.com/sous-chefs/php).
 However, this resource will install the specific apache2 php module package if needed, see the `apache_mod_php_package` helper.
 
-
 **Note: call this resource directly, not through `apache2_module`!**
 This resource will call `_module` with the correct identifiers for you.
 
@@ -15,4 +14,3 @@ This resource will call `_module` with the correct identifiers for you.
 | ------------ | ------ | ------------------------------------------- | ----------------------------------- |
 | module_name  | String | `php#{node['php']['version'].to_i}_module`  | The name of the php module.         |
 | so_filename  | String | `apache_mod_php_filename`                   | The filename of the module object.  |
-
--- a/documentation/resource_apache2_mod_php.md
+++ b/documentation/resource_apache2_mod_php.md
@@ -3,6 +3,8 @@
 Enables apache2 module `mod_php`.
 
 This resource assumes that php has already been installed with the [`php` cookbook](https://github.com/sous-chefs/php).
+However, this resource will install the specific apache2 php module package if needed, see the `apache_mod_php_package` helper.
+
 
 **Note: call this resource directly, not through `apache2_module`!**
 This resource will call `_module` with the correct identifiers for you.
@@ -12,5 +14,5 @@ This resource will call `_module` with the correct identifiers for you.
 | Name         | Type   | Default                                     | Description                         |
 | ------------ | ------ | ------------------------------------------- | ----------------------------------- |
 | module_name  | String | `php#{node['php']['version'].to_i}_module`  | The name of the php module.         |
-| so_filename  | String | `libphp#{node['php']['version'].to_i}.so`   | The filename of the module object.  |
+| so_filename  | String | `apache_mod_php_filename`                   | The filename of the module object.  |
 
--- a/documentation/resource_apache2_mod_php.md
+++ b/documentation/resource_apache2_mod_php.md
@@ -2,8 +2,8 @@
 
 Enables apache2 module `mod_php`.
 
-This resource assumes that php has already been installed with the [`php` cookbook](https://github.com/sous-chefs/php).
-However, this resource will install the specific apache2 php module package if needed, see the `apache_mod_php_package` helper.
+This resource will install and enable the Apache PHP module. See `apache_mod_php_package` for the platform-specific module package.
+If installing PHP outside of this resource (i.e. with the [`php` cookbook](https://github.com/sous-chefs/php)), you should set `install_package` to false to avoid a possible version conflict.
 
 **Note: call this resource directly, not through `apache2_module`!**
 This resource will call `_module` with the correct identifiers for you.
@@ -15,4 +15,4 @@ This resource will call `_module` with the correct identifiers for you.
 | module_name      | String | `apache_mod_php_modulename`  | The name of the php module.                                                                                     |
 | so_filename      | String | `apache_mod_php_filename`    | The filename of the module object.                                                                              |
 | package_name     | String | `apache_mod_php_package`     | The package that contains the PHP module itself, which is sometimes not included with the default PHP packages. |
-| install_package  | Bool   | `false`                      | Whether to install the PHP module package.                                                                      |
+| install_package  | Bool   | `true`                       | Whether to install the PHP module package.                                                                      |
--- a/documentation/resource_apache2_mod_php.md
+++ b/documentation/resource_apache2_mod_php.md
@@ -10,7 +10,9 @@ This resource will call `_module` with the correct identifiers for you.
 
 ## Properties
 
-| Name         | Type   | Default                      | Description                         |
-| ------------ | ------ | ---------------------------- | ----------------------------------- |
-| module_name  | String | `apache_mod_php_modulename`  | The name of the php module.         |
-| so_filename  | String | `apache_mod_php_filename`    | The filename of the module object.  |
+| Name             | Type   | Default                      | Description                                                                                                     |
+| ---------------- | ------ | ---------------------------- | --------------------------------------------------------------------------------------------------------------- |
+| module_name      | String | `apache_mod_php_modulename`  | The name of the php module.                                                                                     |
+| so_filename      | String | `apache_mod_php_filename`    | The filename of the module object.                                                                              |
+| package_name     | String | `apache_mod_php_package`     | The package that contains the PHP module itself, which is sometimes not included with the default PHP packages. |
+| install_package  | Bool   | `false`                      | Whether to install the PHP module package.                                                                      |
--- a/documentation/resource_apache2_mod_php.md
+++ b/documentation/resource_apache2_mod_php.md
@@ -1,0 +1,16 @@
+# apache2_mod_php
+
+Enables apache2 module `mod_php`.
+
+This resource assumes that php has already been installed with the [`php` cookbook](https://github.com/sous-chefs/php).
+
+**Note: call this resource directly, not through `apache2_module`!**
+This resource will call `_module` with the correct identifiers for you.
+
+## Properties
+
+| Name         | Type   | Default                                     | Description                         |
+| ------------ | ------ | ------------------------------------------- | ----------------------------------- |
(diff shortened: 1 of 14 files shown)


kitchen-porter:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Please include a CHANGELOG entry.</td>
    </tr>
  </tbody>
</table>

</p>

libraries/helpers.rb
```diff
@@ -391,6 +392,25 @@ def pagespeed_url
...
+        end
+      end
+
+      def apache_mod_php_filename
+        if platform_family?('debian')
+          "libphp#{node['php']['version'].to_f}.so" # libphp7.2.so
+        else
+          "libphp#{node['php']['version'].to_i}.so" # libphp7.so
+        end
+      end
```
ramereth, line 413: I don't think we want to depend on the php cookbook to figure this out. Instead, this helper method should determine what this will likely be based on the default repositories setup on the various platforms and the versions we support. Users can still override this via the resource using the ``module_name`` parameter if they use other repositories that may pull in different versions.

Xorima: Just to add to this we do not want to depend on any node attributes which are not automatic from ohai

libraries/helpers.rb
```diff
@@ -396,19 +396,32 @@ def default_site_template_source
...
+          else
+            'libphp7.0.so'
+          end
         else
-          "libphp#{node['php']['version'].to_i}.so" # libphp7.so
+          if (platform_family?('rhel') && node['platform_version'].to_i == 7) || platform?('amazon')
+            'libphp5.so'
+          else
+            'libphp7.so'
+          end
```
ramereth: Why not do this like the following instead?
```ruby
when 'rhel', 'amazon'
  if (platform_family?('rhel') && node['platform_version'].to_i == 7) || platform?('amazon')
    'libphp5.so'
  else
    'libphp7.so'
  end
end
```
This will ensure this catches only rhel/amazon based platform families.

resources/mod_php.rb
```diff
@@ -0,0 +1,39 @@
...
+         default: lazy { apache_mod_php_modulename },
+         description: 'Module name for the Apache PHP module.'
+
+property :so_filename, String,
+         default: lazy { apache_mod_php_filename },
+         description: 'Filename for the module executable.'
+
+action :create do
+  # install mod_php package
+  package apache_mod_php_package unless apache_mod_php_package.nil?
```
ramereth, line 14: We really should have two additional parameters for this resource to account for the different ways an admin can install mod_php (especially on CentOS).
1. We should have a ``package_name`` parameter which defaults to ``apache_mod_php_package``. However if a user is using something such as IUS repositories, the name will be slightly different and we should be able to use different names
2. I think we should also have a ``install_package`` true/false (defaults to ``true``) to allow folks who install this via source or using some other means external to this cookbook.

test/cookbooks/test/recipes/php.rb
```diff
@@ -0,0 +1,24 @@
+::Chef::Recipe.include Apache2::Cookbook::Helpers
+
+include_recipe 'php'
```
ramereth, line 3: This isn't needed since we're directly installing it, please remove it.

test/integration/mod_php/controls/default.rb
```diff
@@ -0,0 +1,35 @@
...
+      its('content') { should match %r{SetHandler application/x-httpd-php} }
+    end
+  else
+    describe file('/etc/httpd/mods-available/php.conf') do
+      it { should exist }
+      its('content') { should match %r{SetHandler application/x-httpd-php} }
+    end
+  end
+
+  describe http('localhost/info.php', enable_remote_worker: true) do
```
ramereth, line 31: ``enable_remote_worker`` is enabled by default.
```suggestion
  describe http('localhost/info.php') do
```

resources/mod_php.rb
```diff
@@ -0,0 +1,39 @@
...
+  directory '/var/lib/php/session' do
+    owner 'root'
+    group default_apache_group
+    mode '770'
+  end
+
+  apache2_module 'php' do
+    identifier new_resource.module_name
+    mod_name new_resource.so_filename
+    notifies :reload, 'service[apache2]', :immediately
```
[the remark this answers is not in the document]
detjensrobert: For some reason if it is delayed php does not get enabled correctly.

test/cookbooks/test/recipes/php.rb
detjensrobert: With this removed, the php packages get installed after `_install` cleans up the distro directories, so it does it on the 2nd converge and fails the idempotency check. Should I bump the php suite to 3 converges or leave this in?

ramereth: We should probably implement the same workaround we have in our wrapper cookbook in this one since it will fix that problem. i.e.:
```ruby
directory 'purge distro conf.modules.d' do
  path "#{apache_dir}/conf.modules.d"
  recursive true
  action :nothing # trigger this in other recipes as needed
end

directory 'purge distro conf.d' do
  path "#{apache_dir}/conf.d"
  recursive true
  action :nothing # trigger this in other recipes as needed
end
```

Then we can just notify this resource.

resources/mod_php.rb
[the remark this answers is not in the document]
ramereth: That makes no sense to me. I'll try and take a closer look myself to see what might be going on.

documentation/resource_apache2_mod_php.md
```diff
@@ -0,0 +1,18 @@
...
+This resource will call `_module` with the correct identifiers for you.
+
+## Properties
+
+| Name             | Type   | Default                      | Description                                                                                                     |
+| ---------------- | ------ | ---------------------------- | --------------------------------------------------------------------------------------------------------------- |
+| module_name      | String | `apache_mod_php_modulename`  | The name of the php module.                                                                                     |
+| so_filename      | String | `apache_mod_php_filename`    | The filename of the module object.                                                                              |
+| package_name     | String | `apache_mod_php_package`     | The package that contains the PHP module itself, which is sometimes not included with the default PHP packages. |
+| install_package  | Bool   | `false`                      | Whether to install the PHP module package.                                                                      |
```
ramereth, line 18: IMHO this should be ``true`` as we expect this to work "out of the box".

documentation/resource_apache2_mod_php.md
```diff
@@ -0,0 +1,18 @@
+# apache2_mod_php
+
+Enables apache2 module `mod_php`.
+
+This resource assumes that php has already been installed with the [`php` cookbook](https://github.com/sous-chefs/php).
+However, this resource will install the specific apache2 php module package if needed, see the `apache_mod_php_package` helper.
```
ramereth, line 6: This should be worded differently to reflect that this isn't needed, however if users want to use it with the php cookbook, then can use that cookbook first but set ``install_package`` to ``false.

detjensrobert: The point of this helper was to install the missing module package for the distros where it is not included in the default PHP packages, not to install PHP. I guess it serves that purpose now anyway so I'll update the wording.

detjensrobert: Also, the mod_php package on Suse does not install the rest of PHP. Should I update the package helper to include the php package?

ramereth: This should only be worried about packages that require to get apache working with mod_php. It shouldn't worry about any other package since that can be handled with the php cookbook or by other means. Is the php required to get mod_php to work properly on OpenSUSE?

detjensrobert: Turns out it was something else on SUSE. False alarm.

test/cookbooks/test/metadata.rb
```diff
@@ -5,5 +5,7 @@
 license           'Apache-2.0'
 description       'Testing cookbook for apache2'
 version           '0.0.1'
+
 depends           'apache2'
 depends           'apt'
+depends           'php'
```
ramereth, line 11:
```suggestion
```

test/cookbooks/test/recipes/php.rb
```diff
@@ -0,0 +1,39 @@
...
+  recursive true
+  action :nothing # trigger this in other resources as needed
+end
+
+directory 'purge distro conf.d' do
+  extend Apache2::Cookbook::Helpers
+  path "#{apache_dir}/conf.d"
+  recursive true
+  action :nothing # trigger this in other resources as needed
+end
```
ramereth, line 15: It would be really preferred to have this included in a more global fashion as it can be used in other places in this cookbook. This is troublesome since we've been only doing it in a recipe however we'd need to figure out a way to do this via a library method I believe.

@Xorima @tas50 any ideas on how we can do that in a sane way?

libraries/helpers.rb
```diff
@@ -391,6 +392,57 @@ def pagespeed_url
       def default_site_template_source
         platform_family?('debian') ? "#{default_site_name}.conf.erb" : 'welcome.conf.erb'
       end
+
+      def apache_mod_php_package
+        case node['platform_family']
+        when 'debian'
+          'libapache2-mod-php'
+        when 'rhel', 'fedora'
```
tas50, line 400: amazon?

resources/mod_php.rb
```diff
@@ -0,0 +1,36 @@
+include Apache2::Cookbook::Helpers
+unified_mode true
```
tas50, line 2: We probably don't need this

resources/mod_php.rb
```diff
@@ -0,0 +1,36 @@
...
+
+property :module_name, String,
+         default: lazy { apache_mod_php_modulename },
+         description: 'Module name for the Apache PHP module.'
+
+property :so_filename, String,
+         default: lazy { apache_mod_php_filename },
+         description: 'Filename for the module executable.'
+
+property :package_name, String,
```
tas50, line 12: We should allow this to be a string or array so that users can provide an array of packages if they need to

test/cookbooks/test/recipes/php.rb
```diff
@@ -0,0 +1,39 @@
+::Chef::Recipe.include Apache2::Cookbook::Helpers
+
+directory 'purge distro conf.modules.d' do
+  extend Apache2::Cookbook::Helpers
```
tas50, line 4: This shouldn't be necessary and we need a better way to handle it

test/cookbooks/test/recipes/php.rb
```diff
@@ -0,0 +1,39 @@
...
+end
+
+service 'apache2' do
+  extend Apache2::Cookbook::Helpers
+  service_name lazy { apache_platform_service_name }
+  supports restart: true, status: true, reload: true
+  action [:start, :enable]
+end
+
+apache2_mod_php '' do
```
tas50, line 28: We could allow a nameless resource here by adding this to the resource

```ruby
property :name, String, default: ''
```

test/cookbooks/test/recipes/php.rb
ramereth: So I discussed this with @tas50 today and we decided it's best if we just include this resource/logic directly in the ``apache2_mod_php`` resource so that it will work "out of the box". For other resources we'll just have to replicate this code. Since it's running inside of a resource context, we don't need to worry about duplicated resources.

kitchen.dokken.yml
```diff
@@ -17,7 +17,7 @@ platforms:
     driver:
       mage: dokken/amazonlinux-2
```
ramereth, line 18:
```suggestion
      image: dokken/amazonlinux-2
```

resources/mod_php.rb
```diff
@@ -0,0 +1,49 @@
...
+property :install_package, [true, false],
+         default: true,
+         description: 'Whether to install the Apache PHP module package'
+
+action :create do
+  # install mod_php package (if requested)
+  package new_resource.package_name do
+    only_if { new_resource.install_package }
+    notifies :delete, 'directory[purge distro conf.modules.d]'
+    notifies :delete, 'directory[purge distro conf.d]'
```
ramereth, line 27: We want to ensure these files are removed before we restart apache.
```suggestion
    notifies :delete, 'directory[purge distro conf.modules.d]', :immediately
    notifies :delete, 'directory[purge distro conf.d]', :immediately
```

resources/mod_php.rb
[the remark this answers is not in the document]
ramereth: I believe this is the more correct way to fix this. It should be restarted, not reloaded. It also doesn't need to be immediate.
```suggestion
    notifies :restart, 'service[apache2]'
```

PR #701, opened 2020-08-29
Add tests for package name override

ramereth:

# Description

## Issues Resolved

## Check List

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,6 +54,20 @@ jobs:
           - 'ports'
           - 'module-template'
           - 'mod-php'
+          - 'pkg-name'
+        exclude:
+          - os: debian-9
+            suite: pkg-name
+          - os: debian-10
+            suite: pkg-name
+          - os: centos-8
+            suite: pkg-name
+          - os: ubuntu-1604
+            suite: pkg-name
+          - os: ubuntu-1804
+            suite: pkg-name
+          - os: opensuse-leap-15
+            suite: pkg-name
       fail-fast: false
 
     steps:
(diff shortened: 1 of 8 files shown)


.github/workflows/ci.yml
```diff
@@ -69,6 +83,12 @@ jobs:
         with:
           suite: ${{ matrix.suite }}
           os: ${{ matrix.os }}
+      - name: Print debug output on failure
+        if: failure()
+        run: |
+            set -x
+            sudo journalctl -l --since today
+            KITCHEN_LOCAL_YAML=kitchen.dokken.yml /usr/bin/kitchen exec ${{ matrix.suite }}-${{ matrix.os }} -c "journalctl -l"
```
Xorima, line 91: if we keep doing this everywhere, why don't we extend the github action to do this by default?

resources/mod_ssl.rb
```diff
@@ -1,6 +1,10 @@
 include Apache2::Cookbook::Helpers
 unified_mode true
 
+property :mod_ssl_pkg, String,
+         default: 'mod_ssl',
+         description: ''
```
Xorima, line 6: I think we should have some kind of description, even if it is just this:

```suggestion
         description: 'The name of the mod_ssl package'
```

.github/workflows/ci.yml
ramereth: Should I keep this here for now and work on a separate PR which adds this to the action?

Xorima: yeah, I think that is best.

Xorima (approved):

PR #705, opened 2020-09-12
Directly include Apache2::Cookbook::Helpers in recipes and resources by default

ramereth:
This removes the need to use `extend Apache2::Cookbook::Helpers` in resources
and `::Chef::DSL::Recipe.include Apache2::Cookbook::Helpers` in recipes.

# Description

## Issues Resolved

## Check List

### DIFF
--- a/libraries/helpers.rb
+++ b/libraries/helpers.rb
@@ -1,4 +1,3 @@
-
 module Apache2
   module Cookbook
     module Helpers
@@ -446,3 +445,6 @@ module Apache2
     end
   end
 end
+
+Chef::DSL::Recipe.include Apache2::Cookbook::Helpers
+Chef::Resource.include Apache2::Cookbook::Helpers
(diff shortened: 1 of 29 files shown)


libraries/helpers.rb
```diff
@@ -446,3 +445,6 @@ def apache_mod_php_filename
     end
   end
 end
+
+Chef::Recipe.include Apache2::Cookbook::Helpers
```
tas50, line 449: This is actually the wrong class and I have a cookstyle rule for this, but since you're using the modern .include method it's not triggering. Time to update that. Look at ChefCorrectness/IncorrectLibraryInjection

test/cookbooks/test/recipes/php.rb
```diff
@@ -1,11 +1,8 @@
-::Chef::DSL::Recipe.include Apache2::Cookbook::Helpers
```
tas50, line 1: This is the right class. Chef::DSL::Recipe not Chef::Recipe

tas50:
This will make this all much easier to work with. Thanks @ramereth

axelrtgs (approved):
looks much cleaner to me

Issue #750, opened 2021-06-24
Unable to override options for security and charset configurations with install resource


freakinhippie:
## :ghost: Brief Description

The `conf` resource added the ability to pass in an _:options_ hash, that ultimately is passed into the template as variables. The default value for this hash is:

```ruby
{
  server_tokens: 'Prod',
  server_signature: 'On',
  trace_enable: 'Off',
}
```

These specific settings are used for the `security` template. The security template is called from the `install` resource with no block arguments, making it impossible to override these values idempotently. Further, the `install` resource provides no mechanism to override these default values.

A non-idempotent work-around might look like this:

```ruby
# call the installer
apache2_install 'default'

# overwrite the security.conf template with non-default values
apache2_conf 'security' do
  options(
    server_signature: 'Off',
    server_tokens: 'Full',
    trace_enable: 'Off',
  )
end
```

Unfortunately, this has the effect of restarting the apache service on each chef converge. The same issue is present for the `charset` conf template.

## :pancakes: Cookbook version

8.11.1

## :woman_cook:  Chef-Infra Version

15.x

## :tophat: Platform details

Ubuntu 18.04

## Steps To Reproduce

See above.

## :police_car: Expected behavior

Some way to override the default values.

## :heavy_plus_sign: Additional context

N/A

PR #751, opened 2021-07-08
Allow install resource to override security & charset conf

### LINKED ISSUE #728: ServerName directive absent from apache2.conf.erb
## :ghost: Brief Description

The `config` resource allows for setting the ServerName directive in apache2.conf, as per
https://github.com/sous-chefs/apache2/blob/2a48478d115ba9f156cb10eb0bae0b79fec1d0ec/resources/config.rb#L54-L56

But the default template in `templates/default/apache.conf.erb` does not have a ServerName directive, nor a placeholder for the template variable.

Also telling, the `install` resource does not expose `server_name`, so there's an inconsistency here.

## :pancakes: Cookbook version

HEAD; also using v8.7.0 in my cookbook.

## :woman_cook:  Chef-Infra Version

N/A

## :tophat: Platform details

N/A

## Steps To Reproduce

```ruby
apache2_config 'apache2.conf' do
  server_name 'some_server_name'
end
```
## :police_car: Expected behavior

The ServerName directive would be settable via the resource.

## :heavy_plus_sign: Additional context

N/A

freakinhippie:
* Add `default_charset`, `server_signature`, `server_tokens`, and
`trace_enable` to `install` resource
* Add `install_override` test suite
* Add `ServerName` directive to default `apache2.conf.erb`

# Description

Allows the `install` resource to override the values passed into the `security` and `charset` conf custom resources.

## Issues Resolved

#728 
#750 

## Check List

- [X] All tests pass. See TESTING.md for details.
- [X] New functionality has been documented in the README if applicable.

### DIFF
--- a/templates/security.conf.erb
+++ b/templates/security.conf.erb
@@ -9,7 +9,9 @@
 # Set to one of:  Full | OS | Minimal | Minor | Major | Prod
 # where Full conveys the most information, and Prod the least.
 #
+<% unless @server_tokens.nil? %>
 ServerTokens <%= @server_tokens %>
+<% end %>
 
 #
 # Optionally add a line containing the server version and virtual host
@@ -19,7 +21,9 @@ ServerTokens <%= @server_tokens %>
 # Set to "EMail" to also include a mailto: link to the ServerAdmin.
 # Set to one of:  On | Off | EMail
 #
+<% unless @server_signature.nil? %>
 ServerSignature <%= @server_signature %>
+<% end %>
 
 #
 # Allow TRACE method
@@ -29,4 +33,6 @@ ServerSignature <%= @server_signature %>
 #
 # Set to one of:  On | Off | extended
 #
+<% unless @trace_enable.nil? %>
 TraceEnable <%= @trace_enable %>
+<% end %>
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,6 +56,7 @@ jobs:
           - 'pkg-name'
           - 'ports'
           - 'ssl'
+          - 'install-override'
         exclude:
           - os: debian-9
             suite: pkg-name
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ This file is used to list changes made in each version of the apache2 cookbook.
 
 ## Unreleased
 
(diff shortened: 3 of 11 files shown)


freakinhippie:
The CI does not seem to be running the new test suite. I don't know where to enable that, besides _kitchen.yml_.

ramereth:
@freakinhippie `.github/workflows/ci.yml`

templates/security.conf.erb
```diff
@@ -9,7 +9,9 @@
 # Set to one of:  Full | OS | Minimal | Minor | Major | Prod
 # where Full conveys the most information, and Prod the least.
 #
+<% unless @server_tokens.nil? %>
```
ramereth, line 12: I don't think this is needed at all. When this is set to `nil` this block does not work as expected (same in all places).

freakinhippie: So, remove the `nil` check from each of them?

freakinhippie: The reason that I added those is because the resource can be called directly, with overridden options, and if one of those values is not set than the resulting config would be invalid.

ramereth: Can you please give a code example that would make this happen?

freakinhippie: Sure. The default `options` value for the `conf` resource includes a hash of:

```ruby
{
  server_tokens: 'Prod',
  server_signature: 'On',
  trace_enable: 'Off',
}
```

If the options are overridden with this call:

```ruby
apache2_conf 'security' do
  options(
    server_tokens: 'Full',
    trace_enable: 'Off'
  )
end
```

the `server_signature` value is `nil` and the resulting _security.conf_ file will have the a line like

```
ServerSignature 
```

rather than

```
ServerSignature On
```

freakinhippie: If your preference is that the config _should_ fail if those values are omitted I'm happy to remove the checks.

ramereth: Ah yes, I see what you mean. I was only trying from the `apache2_install` resource, not from the `apache2_conf` resource directly. Thanks for pointing that out.

ramereth:
@freakinhippie can you please address the failure on OpenSUSE? Thanks!

ramereth (approved):

kitchen-porter:
Released as: [8.13.0](https://github.com/sous-chefs/apache2/releases/tag/8.13.0)

PR #800, opened 2023-10-18
Fix issue when using apache2_mod_ssl resource instead a wrapper resource

ramereth:
When doing `apache2_module 'ssl'` instead of another custom resource, the
mod_ssl package does not get installed properly due to the fact it's set to run
in the root context of the chef run. However, it never seems to get installed,
or at least installed in time for it to work leaving a system in a broken state.

I cannot see why this was needed in the first place as it seems to work without
it.

# Description

Describe what this change achieves

## Issues Resolved

List any existing issues this PR resolves

## Check List

- [ ] A summary of changes made is included in the CHANGELOG under `## Unreleased`
- [ ] New functionality has been documented in the README if applicable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ This file is used to list changes made in each version of the apache2 cookbook.
 
 ## Unreleased
 
+- Fix issue when using apache2_mod_ssl resource instead a wrapper resource
+
 ## 9.0.5 - *2023-09-28*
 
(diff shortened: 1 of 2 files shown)


mvangoor (approved):

bmhughes (approved):

kitchen-porter:
Released as: [9.0.6](https://github.com/sous-chefs/apache2/releases/tag/9.0.6)

